Issue #508, opened 2024-07-12
`monkey_run_tour()` cannot reach children steps of `DatasetPermissionStep`


wiitt:
Trying to design a unit test for a case when an additional dataset is provided with `MoreDatasetsStep`, discovered that children steps of `DatasetPermissionStep` cannot be reached with `RecursiveAnswers` in `monkey_run_tour()`.
To my understanding, mechanism of adding children steps for `DatasetPermissionStep` is different from how children steps are added in all other steps. For `DatasetPermissionStep`, children steps are created at the stage of tour initialization, while, for all other steps, children steps are added inside of Step classes. This difference make impossible to reach children steps of `DatasetPermissionStep` with `RecursiveAnswers` since `recursive_helper()` propagates into children steps only when the number of children steps before running the step is smaller than the number of children steps after running the step. As children steps of `DatasetPermissionStep` are added when the tour is initialized (not during running the step), recursive condition is not triggered which causes errors in `monkey_run_tour()`.

Issue #470, opened 2024-06-14
Automatically setting `multispeaker` and `multilingual` should look at all datasets not just whether a specific dataset is multilingual or multispeaker


roedoejet:
We might use the wizard to combine multiple single-speaker, single-language datasets. In doing so, `multilingual` and `multispeaker` never get set to `True` by default, but they should be. 

So, we need:
- meta logic at the end of the wizard process to scan all the data for languages and speakers
- as soon as there is more than one language, `multilingual` should be set to true
- if the speaker is set to `default` in two different languages - the speaker name should actually be set to something different. i.e. the speaker `default` in a single-speaker English dataset is probably a different speaker than `default` in a single-speaker Sinhala dataset.
- if we find more than one speaker across dataset, `multispeaker` should be set to true

PR #509, opened 2024-07-15
fix(tests): removes blocking of reaching children steps

wiitt:
Children steps created at the tour initialization can be reached now. Fixes https://github.com/EveryVoiceTTS/EveryVoice/issues/508

### PR Goal? 
Modifies the condition for the children steps recursion in `monkey_run_tour()`, removes unreasonable blocking.

### Fixes? 
Fixes https://github.com/EveryVoiceTTS/EveryVoice/issues/508

### Feedback sought? 
Are there any cases when we don't want to run children steps of a step which has a number of children steps reduced after its run?

### Priority? 
Medium. Blocks testing of the developed solution for the issue https://github.com/EveryVoiceTTS/EveryVoice/issues/470.

### Tests added? 
Adds no new tests.

### How to test? 
Run wizard testing and check if this PR breaks anything. 

### Confidence? 
Medium. Don't know why this condition was initially added.

### Version change? 
No. Fix of a bug in tests which hasn't revealed itself so far.

### Related PRs? 
No, so far.

### DIFF
--- a/everyvoice/tests/test_wizard.py
+++ b/everyvoice/tests/test_wizard.py
@@ -648,14 +648,9 @@ class WizardTest(TestCase):
             """
             for step_and_answer in steps_and_answers:
                 step = step_and_answer.step
-                # print(step.name)
-                old_children_len = len(step.children)
                 with step_and_answer.monkey:
                     step.run()
-                if (
-                    len(step.children) > old_children_len
-                    and step_and_answer.children_answers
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
I can't think of any situation where this wouldn't be OK. I'm curious why it was written this way in the first place - maybe only step additions were considered?

Issue #476, opened 2024-06-19
The demo should not display an option for language or speaker if there is only one option


roedoejet:
We currently allow dropdowns for language and speaker even if there is only one to choose from. We should dynamically display these fields.

PR #511, opened 2024-07-16
fix(demo): locks dropdown lists when only one option is available

wiitt:
Fixes https://github.com/EveryVoiceTTS/EveryVoice/issues/476

### PR Goal? 
Locks fields in the demo which do not contain options to choose from. 

### Fixes? 
Fixes https://github.com/EveryVoiceTTS/EveryVoice/issues/476.

### Feedback sought? 
It's possible to hide these fields when they are not interactive, but I think it's better to still display them to indicate speaker/language used for model training. What is your opinion?

### Priority? 
Low.

### Tests added? 
No tests are added.

### How to test? 
Run demo with multilingual/multispeaker and monolingual/monospeaker models. Check that fields are locked when its appropriate.

### Confidence? 
High.

### Version change? 
No.

### Related PRs? 
No.

### DIFF
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -2,7 +2,7 @@ import json
 import sys
 from enum import Enum
 from pathlib import Path
-from typing import Any
+from typing import Any, List
 
 import typer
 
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
Yes - this is nice to just use the `interactive` kwarg. good find.

PR #516, opened 2024-07-22
Fix punctuation handling

roedoejet:
### PR Goal? 

Add the validation and punctuation handling discussed in https://github.com/EveryVoiceTTS/EveryVoice/issues/484

### Fixes? 

#450 and #484

### Feedback sought? 

confirmation this is the strategy we want to use

### Priority? 

medium

### Tests added? 

yes

### How to test? 

try running it on an existing config with punctuation symbols in your character list

### Confidence? 

medium

### Version change? 

yes, should trigger a version change as it is breaking

PR #515, opened 2024-07-22
fix: change shape of filelist list data instead of re-reading it

roedoejet:
this allows for any processes that take place on the filelist list data (i.e. normalization) to be preserved

fixes https://github.com/EveryVoiceTTS/EveryVoice/issues/504

### PR Goal? 

Ensures that we are outputting a set of symbols without duplicates

### Fixes? 

#504 

### Feedback sought? 

sanity. code clarity

### Priority? 

medium

### Tests added? 

tests updated

### How to test? 

You could try re-running on your dataset to see how the config changes

### Confidence? 

medium

### Version change? 

none. this isn't a breaking change. no version change needed really.

PR #518, opened 2024-07-24
fix: add whitespace collapsing and text stripping by default

roedoejet:
### PR Goal? 

Whitespace collapsing wasn't being applied by the wizard, only the chosen cleaners. This changes that.

### Feedback sought? 

sanity, code check

### Priority? 

medium

### Tests added? 

✅ 

### Confidence? 

medium-high

### Version change? 

no

### Related PRs? 

#515 and #516

### DIFF
--- a/everyvoice/config/text_config.py
+++ b/everyvoice/config/text_config.py
@@ -5,7 +5,7 @@ from pydantic import BaseModel, ConfigDict, Field, model_validator
 from everyvoice.config.shared_types import ConfigModel
 from everyvoice.config.utils import PossiblySerializedCallable
 from everyvoice.text.utils import normalize_text_helper
-from everyvoice.utils import collapse_whitespace
+from everyvoice.utils import collapse_whitespace, strip_text
 
 
 class Punctuation(BaseModel):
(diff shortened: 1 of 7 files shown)


wiitt (commented):
Is it an expected behaviour that, if no text normalization is applied, character list remains duplicates of empty symbols?
For example, creating a project from `/sgile/data/MohawkCorpus/am_corpus` without text normalization results in
```
moh_characters: ['', '', (, ), '0', '1', '3', '8', a, c, d, e, g, h, i, k, l, m,
    n, o, p, r, s, t, u, v, w, x, y, z, '', à, á, è, é, ì, í, ò, ó]
```
where you can find three instances of ''.

roedoejet:
> Is it an expected behaviour that, if no text normalization is applied, the character list remains duplicates of empty symbols? For example, creating a project from `/sgile/data/MohawkCorpus/am_corpus` without text normalization results in
> 
> ```
> moh_characters: ['', '', (, ), '0', '1', '3', '8', a, c, d, e, g, h, i, k, l, m,
>     n, o, p, r, s, t, u, v, w, x, y, z, '', à, á, è, é, ì, í, ò, ó]
> ```
> 
> where you can find three instances of ''.
> 
> I also noticed that changes from #515 are highlighted as changes of this PR as well.

great find @wiitt - no, this is not intended behaviour. It was an unnecessary `if response:` condition that was preventing it from applying. I've fixed this and added a test to catch this in the future.

PR #524, opened 2024-08-20
fix(deps): bump nltk to 3.9.1 to fix CVE

joanise:
### PR Goal? 

Bump nltk to clear CVE report, although in this case it probably didn't really matter, I don't think we're using the vulnerable code in NLTK.

### Fixes? 

Fixes https://github.com/EveryVoiceTTS/EveryVoice/security/dependabot/7

### Priority? 

low

### Tests added? 

already covered


SamuelLarkin (approved):
LGTM

PR #528, opened 2024-08-27
Fix the pre-commit CI workflow

joanise:
### PR Goal? 

pre-commit CI, because of mypy, needs a full everyvoice install, with `pip install -e .`

fix that, and fix one mypy error that was tripping mypy in CI


SamuelLarkin (approved):
It's alive!

PR #529, opened 2024-08-27
feat(ci): Merged pre-commit into test workflow

SamuelLarkin:
### PR Goal? 

Merge pre-commit workflow into the test workflow since both of the require to build an environment and install EveryVoice which is a bit expensive.

### Fixes? 

None

### Feedback sought? 

Merging approval and to validate the step order.
Should we run the pre-commit hooks before or after the testsuite?

### Priority? 

Low

### Tests added? 

None

### How to test? 

Look at the github actions.

### Confidence? 

Good

### Version change? 

No

### Related PRs? 

#528 

Requested by @joanise


.github/workflows/test.yml
```diff
@@ -35,13 +35,19 @@ jobs:
...
-      - name: Log complete list of packages
-        run: |
-          pip freeze
+      - run: pip freeze
+      - run: pip list
       - name: Check licenses
         run: |
           pip install pip-licenses
           if pip-licenses | grep -E -v 'Artistic License|LGPL|Public Domain' | grep GNU; then echo 'Please avoid introducing *GPL dependencies'; false; fi
+      - uses: tj-actions/changed-files@v36
```
joanise, line 44: `tj-actions/changed-files@v36` is giving us a node 16 warning (see https://github.com/EveryVoiceTTS/EveryVoice/actions/runs/10581973532)
any reason not to use the current `@v45`?

joanise (commented):
Looks almost good to me. I think we should bump the changed-files to clear the warning, and I guess I would like to know what happens when a file is not already black or isort compliant. Does it get committed by pre-commit-ci/lite-action? What happens when mypy fails?

.github/workflows/test.yml
SamuelLarkin: I'll update to v45

SamuelLarkin: I'll also try out a not properly blacked file.  I guess I'll have to use nano because my ide auto formats.

.github/workflows/test.yml
```diff
@@ -32,16 +34,23 @@ jobs:
         run: |
           CUDA_TAG=cpu pip install -r requirements.torch.txt --find-links https://download.pytorch.org/whl/torch_stable.html
           pip install -r requirements.test.txt
```
joanise, line 36: Oops, I thought we had removed this `requirements.test.txt` since dev includes it.

joanise (approved):
Nice, after much pair programming, we have it working as desired! 🎉

.github/workflows/test.yml
SamuelLarkin: I also noticed it after and change it to `dev`.

SamuelLarkin:
A BIG thanks to @joanise for helping solving the mystery of `conda-incubator/setup-miniconda` and `pre-commit/action`.

PR #533, opened 2024-08-29
Make sure ffmpeg is installed, and show a clear message if not

### LINKED ISSUE #527: Check ffmpeg installation in `make-everyvoice-env`
Since EveryVoice requires ffmpeg to run the demo, we should check that it is installed and install it otherwise.

joanise:
### PR Goal? 

Make sure ffmpeg is installed, and error out for a clear message if not

two changes:
 - `make-everyvoice-env` now just always installs ffmpeg
 - `everyvoice demo` exits right away if ffmpeg is not found, since we know we'll crash when the user eventually clicks on `synthesize`

### Fixes? 

Fixed #527

### Feedback sought? 

general validation

### Priority? 

beta

### Tests added? 

no, because we don't have any CI for `everyvoice demo` at the moment.

### How to test? 

1) get yourself an environment where ffmpeg is not installed (`which ffmpeg` will return nothing) and run `everyvoice demo fp.ckpt vocoder.ckpt` and see it exit with clear instructions.

2) create a new environment with `make-everyvoice-env` and see that `which ffmpeg` is right in your conda env.

### Confidence? 

high

I tested this thoroughly, with ffmpeg not on my path, and with ffmpeg on my path but corrupted (and thus not runnable) and you get the error message each time, whereas with a good ffmpeg things proceed normally.

### Version change? 

no

### DIFF
--- a/everyvoice/demo/app.py
+++ b/everyvoice/demo/app.py
@@ -1,8 +1,11 @@
 import os
+import subprocess
+import sys
 from functools import partial
 
 import gradio as gr
 import torch
+from loguru import logger
 
 from everyvoice.config.type_definitions import TargetTrainingTextRepresentationLevel
 from everyvoice.model.feature_prediction.FastSpeech2_lightning.fs2.cli.synthesize import (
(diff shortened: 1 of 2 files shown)


everyvoice/demo/app.py
```diff
@@ -82,6 +85,17 @@ def synthesize_audio(
     return sr, wav[0]
 
 
+def require_ffmpeg():
+    """Make sure ffmpeg is found and can be run, or else exit"""
+    try:
+        subprocess.run(["ffmpeg", "-h"], capture_output=True)
+    except Exception:
+        logger.error(
+            "ffmpeg not found or cannot be executed.\nffmpeg is required to run the demo.\nPlease install it, e.g., with 'conda install ffmpeg' or with your OS's package manager."
```
roedoejet, line 94: muuuuch better error message :)

roedoejet (approved):
looks good to me, thanks @joanise - much better error messaging

PR #530, opened 2024-08-28
Fix several problems I encountered trying EV from scratch

joanise:
* 2024-08-28 55640a637 Eric Joanis fix: make link about NFC norm visible even on narrow screens
* 2024-08-28 c8f36da85 Eric Joanis fix: wizard should not add '' to the character set
* 2024-08-28 8012a73d9 Eric Joanis feat: give the user early feedback before we load expensive stuff
* 2024-08-28 2e2e16dc1 Eric Joanis fix: always install ffmpeg in make-everyvoice-env
* 2024-08-28 8e850b7a4 Eric Joanis fix: make wizard prompt code clearer and fix crash

### PR Goal? 

Fix a bunch of problems I faced trying to use EV again.

### Fixes? 

Fixes #527
The other problem had an associated issue

### Feedback sought? 

standard code review stuff

### Priority? 

Pretty high because at the moment, the wizard is broken, adding `''` to `test_characters` in `config/everyvoice-shared-text.yaml`, which cause a failure when you run `everyvoice train text-to-spec`~`spec-to-wav`~.

### Tests added? 

most of this was already covered before, I added bits missing

### How to test? 

run the wizard and see:
 - the URL for NFC normalization is now fully on your screen (used to get truncated unless you screen was very wide)
 - `test_characters` no longer contains `''`
 - no longer get a crash when you pick no text transformations

make a new env with `make-everyvoice-env` and see ffmpeg is in the conda env

run `everyvoice train` or `everyvoice preprocess` and see 1 to 3 lines of INFO telling you stuff is loading, so you're not wondering if the command is running or not

### Confidence? 

pretty high

### Version change? 

no

### Related PRs? 

none

PR #535, opened 2024-08-29
multiple patches to the new-project wizard

joanise:
### PR Goal? 

Multiple improvements and fixes to the new-project wizard.

### Fixes?

Fix the problem where the wizard would include `''` in `test_characters` and thus make `everyvoice train text-to-spec` fail.

Fix the problem where the wizard crashes if you select no text transformations.

Make sure the URL for documentation of NFC normalization is fully legible even when your terminal is not super wide.

Other side improvements to the wizard at the same time.

### Feedback sought? 

regular review

### Priority? 

High because at the moment, the wizard is broken, 
 - adding `''` to `test_characters` in `config/everyvoice-shared-text.yaml`, which causes a failure when you run `everyvoice train text-to-spec`;
 - crashing if you select no text transformations.

### Tests added? 

yup

### How to test? 

run the wizard and see:
 - the URL for NFC normalization is now fully on your screen (used to get truncated unless you screen was very wide)
 - `test_characters` no longer contains `''`
 - no longer get a crash when you pick no text transformations

### Confidence? 

high

### Version change? 

no

### Related PRs? 

This was part of #530 but it was too big so I spliced it out, so different work can be reviewed separately.

### DIFF
--- a/everyvoice/wizard/prompts.py
+++ b/everyvoice/wizard/prompts.py
@@ -1,5 +1,5 @@
 import sys
-from typing import Union
+from typing import Iterable, Sequence
 
 import simple_term_menu
 from questionary import Style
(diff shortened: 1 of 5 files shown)


everyvoice/wizard/prompts.py
```diff
@@ -56,12 +65,17 @@ def get_response_from_menu_prompt(
...
+        else:
+            return [choices[i] for i in selection]
     else:
-        if isinstance(index, tuple):
-            return [choices[i] for i in index]
+        if return_indices:
+            return selection
         else:
-            return choices[index]
+            return choices[selection]
```
SamuelLarkin, line 81: Just for my understanding, when `multi` you can select nothing but otherwise you have to select something?

SamuelLarkin (approved):
LGTM

everyvoice/wizard/prompts.py
joanise: Exactly.

PR #536, opened 2024-08-29
Display "Loading x" messages as the EV apps start

joanise:
### PR Goal? 

Currently, when you start a number of EV commands, there's nothing on screen for quite a while with no feedback telling the user we're starting to work on it.

This PR provides helpful "Loading xyz" messages at several key points where the delays can be long.

Affected:
 - `everyvoice demo`
 - `everyvoice train`
 - `everyvoice synthesize`
 - `everyvoice preprocess`
 - and every other command that ends up importing `everyvoice.base_cli.helpers`

### Feedback sought? 

If find these messages helpful. (I really do!)

### Priority? 

low

### Tests added? 

Except for app.py, this was all previously covered, and it seems unnecessary assert anything about info messages.

### How to test? 

Run the commands listed above and see the new info Loading messages.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/85

### DIFF
--- a/everyvoice/base_cli/helpers.py
+++ b/everyvoice/base_cli/helpers.py
@@ -20,6 +20,10 @@ from loguru import logger
 from pydantic import ValidationError
 from tqdm import tqdm
 
+# This is where we start loading expensive stuff for most CLI commands: give the
+# user a signal that's immediately visible so they know something's happening.
+logger.info("Loading everyvoice package and dependencies")
+
 from everyvoice.config.type_definitions import TargetTrainingTextRepresentationLevel
 from everyvoice.exceptions import InvalidConfiguration
 from everyvoice.model.aligner.config import DFAlignerConfig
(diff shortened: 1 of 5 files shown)


joanise:
@roedoejet commented on Slack:
> I’m not sure I agree with parts of 536 - did you measure how long the imports that you’re logging actually take? It seems like overkill to log the loading of `everyvoice` and clutter the terminal with possibly multiple loading messages. I also don’t know if it’s a good idea to log this every time the module is loaded, even programmatically. This seems like something that only matters for the CLI, in which case I think something like a spinner is more appropriate: https://github.com/pavdmyt/yaspin

I'm going to review redoing this PR using yasmin but at first glance what bothers me is that we have a lot of CLI entry points that would need to be changed, it contrast with just a few expensive centralized import points. To be continued but not high priority.

SamuelLarkin (approved):
LGTM

PR #541, opened 2024-09-05
chore: dependency maintenance to allow g2p 2.1

joanise:
also:
 - bump matplotlib to the current version, already bundled with type hints
 - declare numpy < 2 since torch < 2.4.1 requires it without declaring it

### PR Goal? 

Allow g2p 2.1.0 with EveryVoice

### Priority? 

low

### How to test? 

CI test suites should do

### Confidence? 

high


roedoejet (approved):
thanks @joanise !

Issue #534, opened 2024-08-29
Resuming training for text-to-spec model does not resume with appropriate losses


roedoejet:
To recreate: train a text-to-spec model, then add a `finetune_ckpt` value and resume training (you might need to change another model value like the max number of steps). the losses do not resume properly.

Maybe related to https://github.com/EveryVoiceTTS/EveryVoice/pull/473

SamuelLarkin:
# Observation
First run shows some overlap between the previous run (cyan) and resuming (pink)
![image](https://github.com/user-attachments/assets/b7847c1e-eb09-4fb9-993b-d0ff6fc2bc90)

SamuelLarkin:
![image](https://github.com/user-attachments/assets/9bed23c0-fede-46c2-a9ad-a1d140768ec8)
![image](https://github.com/user-attachments/assets/c966bce5-9b99-43c7-87cd-621f96a1087a)
![image](https://github.com/user-attachments/assets/759575df-ce1f-4bdb-a7bf-a1b7fefe43d4)

SamuelLarkin:
# Test Code
```sh
#!/bin/bash
# vim:nowrap:
#
#SBATCH --job-name=resume-EV
#SBATCH --partition=gpu_a100
#SBATCH --account=nrc_ict__gpu_a100
#SBATCH --gres=gpu:1
##SBATCH --partition=standard
##SBATCH --account=nrc_ict
#SBATCH --qos=low
#SBATCH --time=720

#SBATCH --nodes=1
#SBATCH --ntasks-per-node=1
#SBATCH --cpus-per-task=32
#SBATCH --mem=16G

##SBATCH --comment="image=nrc/nrc_all_default_ubuntu-22.04-amd64_latest"
##SBATCH --comment="image=registry.maze-c.collab.science.gc.ca/sschpcs/generic-job:ubuntu22.04_master"

#SBATCH --output=%x-%j.out
#SBATCH --error=%x-%j.err
#SBATCH --open-mode=append
##SBATCH --mail-user==sam037
##SBATCH --mail-type=NONE

#SBATCH --signal=B:15@30

## Priority:
#$ low
## Resources:
#
#

function downsize {
   # Given a new-project, change some default values to make everyvoice faster to test.
   sed \
      -i \
      -e '/use_postnet/ s/: .*/: false/' \
      -e '/batch_size/ s/: .*/: 5/' \
[...]

PR #547, opened 2024-09-12
Resume at the end of the last trained epoch

SamuelLarkin:
### PR Goal? 

Fix proper resuming of `text-to-spec` training.
The state at the end of the last epoch wasn't saved and resuming would be performed from the last saved checkpoint that was the last checkpoint used for validation.  This was producing staggered runs as shown in `tensorboard`.

### Fixes? 

#534 

### Feedback sought? 

merge approval

### Priority? 

low

### Tests added? 

None

### How to test? 

```sh
   srun everyvoice train text-to-spec \
      config/everyvoice-text-to-spec.yaml \
      --config-args training.max_epochs=1 \
```

Check the state of the loops
```sh
python -c 'import torch; import json; m = torch.load("logs_and_checkpoints/FeaturePredictionExperiment/save_on_train_epoch_end/checkpoints/last.ckpt", map_location=torch.device("cpu")); print(json.dumps(m["loops"]["fit_loop"]["epoch_loop.batch_progress"], indent=2))'
```
Which will yield something like the following.  You want to look at `current`'s values.  This run used 11790 training examples split across batches of 16 examples thus, one epoch is 11790/16 ~ 736 batches per epoch.  If, instead, we see 500, the default `val_check_interval`, this would mean that we didn't save at the end of the epoch.
```json
{
  "total": {
    "ready": 4421,
    "completed": 4421,
    "started": 4421,
    "processed": 4421
  },
  "current": {
    "ready": 736,
    "completed": 736,
    "started": 736,
    "processed": 736
  },
  "is_last_batch": true
}
```

Try resuming for a second epoch.
```sh
   srun everyvoice train text-to-spec \
      config/everyvoice-text-to-spec.yaml \
      --config-args training.finetune_checkpoint="logs_and_checkpoints/FeaturePredictionExperiment/base/checkpoints/last.ckpt" \
      --config-args training.max_epochs=2 \
```

Use `tensorboard` and check that the second run's training is NOT staggered with your first run.

```sh
tensorboard --port=2024 --logdir=logs_and_checkpoints  --bind_all
```

### Confidence? 

Good

### Version change? 

No

### Related PRs? 

None

### DIFF
--- a/everyvoice/base_cli/helpers.py
+++ b/everyvoice/base_cli/helpers.py
@@ -243,6 +243,7 @@ def train_base_command(
         every_n_train_steps=config.training.ckpt_steps,
         every_n_epochs=config.training.ckpt_epochs,
         enable_version_counter=True,
+        save_on_train_epoch_end=True,
     )
     # This callback will only save the top-k checkpoints
     # based on minimization of the monitored loss


marctessier:
Yes , confirming that the fin-tune checkpoint it resuming from the end of the previous run.  ( 50 steps ahead)  VS how it was definitely overlapping before.  


I will open a new ticket for the 50 steps ahead but will close this since it is now resolved.  :-)

marctessier (approved):
Look good Samuel.

PR #553, opened 2024-09-20
Dev.ej/clarify text to spec

joanise:
### PR Goal? 

It always takes me too long to figure out which model I need to provide when the help says I need a text-to-spec model.

This PR clarifies that.

### Fixes? 

My lack of intuition for why "spec" and "features" are synonyms... I'm probably not the only one!

### Priority? 

low

### Tests added? 

n/a

### How to test? 

run 
 - `everyvoice demo -h`
 - `everyvoice synthesize from-text -h`

and see the model path description clarified

### Confidence? 

high, but there is one question mark: are all text-to-spec models feature prediction models? If not, and there are other kinds, I would write `(e.g., feature prediction)` instead of `(i.e., feature prediction)`.

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/89

### DIFF
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -441,7 +441,7 @@ def demo(
         file_okay=True,
         exists=True,
         dir_okay=False,
-        help="The path to a trained text-to-spec EveryVoice model.",
+        help="The path to a trained text-to-spec (i.e., feature prediction) EveryVoice model.",
         autocompletion=complete_path,
     ),
     spec_to_wav_model: Path = typer.Argument(


.github/workflows/test.yml
```diff
@@ -45,6 +45,10 @@ jobs:
           if pip-licenses | grep -E -v 'Artistic License|LGPL|Public Domain' | grep GNU; then echo 'Please avoid introducing *GPL dependencies'; false; fi
       - uses: tj-actions/changed-files@v45
         id: file_changes
+        with:
+          # Only run pre-commit on EV files themselves, not on submodules
+          # See https://github.com/EveryVoiceTTS/EveryVoice/issues/555
+          exclude_submodules: true
```
SamuelLarkin, line 51: I'm going to be ask that this be made a separate commit / PR as this is not addressing the `text-to-spec` confusion but rather arose from fixing the confusion.

SamuelLarkin (changes requested):
I prefer PR/commits that are not mixing 2 issues.

.github/workflows/test.yml
joanise: CI fixes separated into #557

roedoejet (approved):
lgtm

Issue #439, opened 2024-05-27
Add ReadAlong export format


roedoejet:
See https://github.com/roedoejet/EveryVoice/pull/436 for a similar PR that adds support for exporting TextGrids.

PR #556, opened 2024-09-20
A bunch of small improvements while working on the wizard back and resume

joanise:
### PR Goal? 

I made these changes while working on the wizard's back and resume functions, but I want to PR them separately because they will be confusing to review as part of that PR.

### Fixes? 

 - remove irrelevant unit-testing code from the Step class
 - make more test suites compatible with Windows
 - don't let the wizard leave directories around just to test that we're able to create them
 - turn the warning about not running the sox effects on Windows into a fatal error, unless an explicit environment variable is set: we never want this warning to be ignorable!

### Feedback sought? 

sanity check

### Priority? 

medium: these changes will be required for the wizard improvement PR

### Tests added? 

yes

### How to test? 

 - Should you really care to do so (but don't!), run the test suites on Windows, and see fewer failures.
 - Tell the wizard to save files in a deep directory that doesn't exist yet, then quit before you're done. Before this PR, most of the deep directory is left behind, with this PR you can see it is not.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

n/a

### DIFF
--- a/everyvoice/tests/test_configs.py
+++ b/everyvoice/tests/test_configs.py
@@ -1,12 +1,19 @@
 import json
-import tempfile
 import time
 from pathlib import Path
+from tempfile import TemporaryDirectory
 from typing import Callable
 from unittest import TestCase
 
 import yaml
 from pydantic import ValidationError
+
+try:
+    from readalongs.portable_tempfile import (
+        PortableNamedTemporaryFile as NamedTemporaryFile,
+    )
+except ImportError:
+    from tempfile import NamedTemporaryFile
 
 from everyvoice import exceptions
 from everyvoice.config.preprocessing_config import (
@@ -91,7 +98,7 @@ class ConfigTest(BasicTestCase):
         self.assertEqual(config_32.training.batch_size, 32)
 
     def test_config_save_dirs(self):
-        with tempfile.TemporaryDirectory(prefix="test_config_save_dirs") as tempdir:
+        with TemporaryDirectory(prefix="test_config_save_dirs") as tempdir:
             # Preprocessing Config
             tempdir = Path(tempdir)
             with self.assertRaises(ValidationError):
@@ -101,7 +108,7 @@ class ConfigTest(BasicTestCase):
             self.assertTrue((tempdir / preprocessing_config.save_dir).exists())
 
     def test_config_partial(self):
-        with tempfile.TemporaryDirectory(
+        with TemporaryDirectory(
             prefix="test_config_partial"
         ) as tempdir_str, init_context({"writing_config": Path(tempdir_str)}):
             # Preprocessing Config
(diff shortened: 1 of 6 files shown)


everyvoice/tests/test_configs.py
```diff
@@ -1,13 +1,20 @@
...
 import yaml
 from pydantic import ValidationError
 
+try:
+    from readalongs.portable_tempfile import (
+        PortableNamedTemporaryFile as NamedTemporaryFile,
+    )
+except ImportError:
+    from tempfile import NamedTemporaryFile
+
```
SamuelLarkin, line 17: Is there a way not to require readalongs?  Is EV already dependent on readalongs?  Wouldn't it be wiser to have that `readalongs.protable_temfile` be a new library that both EV and readalongs could import?

everyvoice/wizard/__init__.py
```diff
@@ -159,18 +150,28 @@ def add_steps(self, steps: Sequence[Step | list[Step]], parent: Step):
...
                 self.add_step(item, parent)
 
-    def add_step(self, step: Step, parent: Step, child_index=0):
+    def add_step(self, step: Step, parent: Step):
+        """Insert a step in the specified position in the tour.
+
+        Args:
+            step: The step to add
+            parent: The parent to add the step to
+        """
```
SamuelLarkin, line 159: 👍🏼 Thank you for adding documentation for the methods.

everyvoice/wizard/basic.py
```diff
@@ -141,6 +141,27 @@ def sanitize_input(self, response):
         response = super().sanitize_input(response)
         return response.strip()
 
+    def can_mkdir(self, path: Path) -> bool:
+        """Make sure it's possible to create path, without leaving it behind."""
+        dirs_to_make = []
+        d = path
+        while not d.exists() and d not in (Path("/"), Path(""), Path(".")):
+            dirs_to_make.append(d)
+            d = d.parent
```
SamuelLarkin, line 150: Will this survive `path=="/somewhere_exist/doesntexist/doesnexists/../../a/b/c`?
I guess it depends on `d.parent`.

SamuelLarkin (approved):
I would prefer not to import readalongs.
I have a question about `can_make_dir()`.

everyvoice/tests/test_configs.py
joanise: We don't depend on readalongs yet, and with the try/except we still don't really, which is why I did not add it to requirements.

Also, when we do #439, we will then really add the readalongs dependency to support that export feature, which is in part why I thought I could slide this weird thing in.

everyvoice/wizard/basic.py
joanise: Yes, this is exactly the use case I was working to support, and the reason I did this patch. Previously, validation would have left `/somewhere_exist/doesntexist/doesnexists/../../a/b` on the filesystem, now this while loop will add all the components to `dirs_to_make` and stop at `/somewhere_exist`, since it exists.

joanise: And the finally block at the end of this function will remove them all so we leave no traces behind just for trying to validate that we can create this deep directory.

PR #561, opened 2024-09-26
Dev.ej/wizard navigation

### LINKED ISSUE #468: wizard: Saying yes, there is a speaker/language column when there isn't one gets you in a dead end
If your data does not have either a speaker or a language column, but you say yes to that question, you get painted in a corner and you cannot recover. We should add a way to backtrack.

### LINKED ISSUE #545: Wizard navigation and progress saving
Spliced out of #7

- [ ] there is no "back button" if you make a mistake
- [ ] if you exit out of the config wizard, there is no way to save your progress. It might be nice to think about how to do something like that. I.e. some sort of declarative syntax for how questions would get answered that could double as a progress saver and as a testing aide.

### LINKED ISSUE #546: Question "title"s get truncated in the wizard
In `get_response_from_menu_prompt()`, the `prompt_text` gets wrapped correctly into a Panel,  but the `title` we pass to `simple_term_menu`, is truncated at the edge of the terminal windows.

joanise:
### PR Goal? 

Add several navigation features to the EveryVoice new-project wizard:
 - The ability to go back to a previous step (use Ctrl-C, option 1)
 - The ability to view the question tree (Ctrl-C, option 3, with hidden option `--trace` to auto display it at each step)
 - Hidden option `--debug-state` displays the current state before each question
 - The ability to save progress (Ctrl-C, option 4), and the `--resume-from` option to resume from a saved state

### Fixes? 

Fixes #545
Fixes #468
Fixes #546

### Feedback sought? 

Code review, and playing around with the wizard to see how it feels, how it works, improvements you might like to see.

### Priority? 

beta

### Tests added? 

~~Not yet - I'm making this draft PR to seek feedback but I'll add unit testing before finalizing it.~~

In progress: navigation has unit testing now.

### How to test? 

Run the wizard with the new options, hit Ctrl-C are various points during the wizard and play with the options.

### Confidence? 

medium-high, since I have tested it interactively quite extensively, but will be higher once unit testing is in place.

### Version change? 

might have warranted a minor version bump since it's new features, but we're preparing the next alpha so that doesn't make sense. Nothing broken, in any case, so no major bump.

### Related PRs? 

nope, it's all in EV itself.


marctessier (approved):
Looks good Eric , nice work!   Super practical what you did!   I ran a series of various tests, all with success :-)

PR #563, opened 2024-10-07
Add a yaspin spinner on CLI commands while expensive imports are happening

joanise:
### PR Goal? 

When you run many EveryVoice commands, there's no feedback on screen telling you the command is doing something, and isn't just hanging. With this PR, expensive imports get a spinner that looks like `⠼ Loading... (0:00:10.61)` while loading, and `Done Loading (0:00:21.57)` when done loading.

If the process is run interactively, update is every 9th of a second, but if it's going to a log it's only every 2 seconds, to avoid filling up the log file with garbage.

### Fixes? 

My impatience waiting for stuff to actually happen when I run EV commands.

### Feedback sought? 

How you like it.

### Priority? 

low

### Tests added? 

no, we don't have the framework to test most of these yet, athough some of these commands will be covered by existing unit testing, and I don't feel the need to add assertions, all that matters is that things continue to run as they used to.

### How to test? 

First, run `pip install yaspin`, since that dependency was just added by this PR.

Then run EV train, synthesize, preprocess, etc. Ditto fs2l, hfgl and dfaligner commands.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/91
https://github.com/EveryVoiceTTS/HiFiGAN_iSTFT_lightning/pull/36
https://github.com/EveryVoiceTTS/DeepForcedAligner/pull/25

### DIFF
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -31,6 +31,7 @@ from everyvoice.model.vocoder.HiFiGAN_iSTFT_lightning.hfgl.cli import (
     synthesize as synthesize_hfg,
 )
 from everyvoice.model.vocoder.HiFiGAN_iSTFT_lightning.hfgl.cli import train as train_hfg
+from everyvoice.utils import generic_psv_filelist_reader, spinner
 from everyvoice.wizard import (
     ALIGNER_CONFIG_FILENAME_PREFIX,
     PREPROCESSING_CONFIG_FILENAME_PREFIX,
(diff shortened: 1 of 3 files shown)


marctessier (approved):
Looks good , no issues.   It is especially nice to see the "spinner"  when doing  a synthesize on the head node where it takes a while to load up everything...( over 20sec...)

roedoejet (approved):
I didn't test it, but the code looks good - thanks for doing this @joanise ! I'm happy with this implementation

PR #569, opened 2024-10-28
fix: display less confusing reports in preprocessing

### LINKED ISSUE #342: Fix reporting in preprocessing
Currently we print the report twice because we print a partial report [here](https://github.com/roedoejet/EveryVoice/blob/d117ece5d29947e171bf181bc481c66bb26352ff/everyvoice/preprocessor/preprocessor.py#L769C21-L772C39) and then later a full report [at the end](https://github.com/roedoejet/EveryVoice/blob/d117ece5d29947e171bf181bc481c66bb26352ff/everyvoice/preprocessor/preprocessor.py#L824). We should still print the partial report related to the audio, but we should remove all non-audio related fields, and we should print it in a panel that mentions that it is a partial report.

The summary.txt should only be printed at the end.

Marc would also like to see a list of the files in the summary.txt that got skipped.

joanise:
### PR Goal? 

Reduce the confusion from the two nearly identical reports printed during `everyvoice preprocess`.

### Fixes? 

Fixes #342

### Priority? 

beta

### Tests added? 

already covered by existing tests

### How to test? 

Run `everyvoice preprocess config/everyvoice-text-to-spec.yaml` and see:
 - The first report is labelled partial, and omits missing symbols, skipped processes, and nans, which are not related to audio-processing.
 - The second report is like before, with this elements included.
 
### Confidence? 

medium-high

### Question

The `nans` line in the report is never actually populated. What's it supposed to count? should `_interpolate()` or `extract_pitch()` update it?

### DIFF
--- a/everyvoice/preprocessor/preprocessor.py
+++ b/everyvoice/preprocessor/preprocessor.py
@@ -325,27 +325,35 @@ class Preprocessor:
         seconds %= 60
         return f"{hours}h {minutes}m {seconds}s"
 
-    def report(self, tablefmt="simple"):
+    def report(self, tablefmt="simple", audio_only=False):
         """Print a report of the dataset processing"""
         headers = ["type", "quantity"]
-        table = [
+        audio_1 = [
             ["processed files", self.counters.value("processed_files")],
             [
                 "previously processed files",
                 self.counters.value("previously_processed_files"),
             ],
             ["missing files", self.counters.value("missing_files")],
+        ]
+        others = [
             [
                 "missing symbols",
                 len(self.text_processor.missing_symbols) if self.text_processor else 0,
             ],
             ["skipped processes", self.counters.value("skipped_processes")],
             ["nans", self.counters.value("nans")],
+        ]
+        audio_2 = [
             ["audio_empty", self.counters.value("audio_empty")],
             ["audio_too_short", self.counters.value("audio_too_short")],
             ["audio_too_long", self.counters.value("audio_too_long")],
             ["duration", self.print_duration()],
         ]
+        if audio_only:
+            table = audio_1 + audio_2
+        else:
+            table = audio_1 + others + audio_2
         return tabulate(table, headers, tablefmt=tablefmt)
 
     def get_speaker_and_language(self, item):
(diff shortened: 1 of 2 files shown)


everyvoice/preprocessor/preprocessor.py
```diff
@@ -325,27 +325,35 @@ def print_duration(self):
...
+        audio_2 = [
             ["audio_empty", self.counters.value("audio_empty")],
             ["audio_too_short", self.counters.value("audio_too_short")],
             ["audio_too_long", self.counters.value("audio_too_long")],
             ["duration", self.print_duration()],
         ]
+        if audio_only:
+            table = audio_1 + audio_2
+        else:
+            table = audio_1 + others + audio_2
```
joanise, line 356: Question: should the other statistics be placed after the audio ones, instead of between audio_1 and audio_2?

everyvoice/tests/test_wizard.py
```diff
@@ -755,9 +755,9 @@ def test_wrong_fileformat_festival(self):
         with patch_menu_prompt((0, 1, 2, 3), multi=True) as stdout:
             format_step.run()
         output = stdout.getvalue().replace(" \n", " ")
-        self.assertRegex(output, "does not look like a .*'psv'")
-        self.assertRegex(output, "does not look like a .*'tsv'")
-        self.assertRegex(output, "does not look like a .*'csv'")
+        self.assertRegex(output, r"does not look like a .*'psv'".replace(" ", r"\s"))
+        self.assertRegex(output, r"does not look like a .*'tsv'".replace(" ", r"\s"))
+        self.assertRegex(output, r"does not look like a .*'csv'".replace(" ", r"\s"))
```
joanise, line 760: These failed due to my specific screen size and working path (there was a newline between "does not" and "look like"), so I made them more robust.
Not related to this PR, I know, but I had to do it to run the unit tests.

marctessier (approved):
Reports is no longer displaying the "none" audio fields.

below removed.
```
missing symbols             0                                                            
skipped processes           0                                                            
nans                        0      
```

I will the approve the changes to have this merged into main  and out of the way  and open  a separate issue for having maybe a special reported added to the output log that could generate a list of missing audio files from the given filelist.

PR #578, opened 2024-11-05
Dev.ej/460

### LINKED ISSUE #460: forward slash in the text causes a cryptic error
When there is a forward slash in the text, it causes the tokenization to crash, e.g.:

```
OutOfVocabularySymbolError: Sequence ['k', 'a', 'i', 'e', 's', 'h', 'ó', 'n', 'h', 't', 's', 'h', 'e', 'r', 'a', "'", '', '', 'k', 'a', 's', 't', 'e', 
'r', 'i', 's', 't', 'á', 'h', 't', 's', 'h', 'e', 'r', 'a', "'", ' ', 't', 'ó', 'k', 'a', "'", ' ', 'n', 'ò', 'n', ':', 'w', 'a', ' ', 'e', 'n', 'w', 
'á', 'h', 's', 'e', 'h', 't', 'e', "'", ' ', 'n', 'e', ' ', 'k', 'a', "'", 'n', 'i', 'k', 'o', 'n', 'h', 'r', 'a', 'k', 's', 'é', 'n', 'h', 't', 's', 
'h', 'e', 'r', 'a', "'", '.'] contains item 
```

### LINKED ISSUE #540: An input breaks the preprocess pipeline
The input is:
`kaieshónhtshera'/kasteristáhtshera'` which is in our dataset

Which will lead to
`everyvoice.exceptions.OutOfVocabularySymbolError: Sequence ['k', 'a', 'i', 'e', 's', 'h', 'ó', 'n', 'h', 't', 's', 'h', 'e', 'r', 'a', "'", '', '', 'k', 'a', 's', 't',
'e', 'r', 'i', 's', 't', 'á', 'h', 't', 's', 'h', 'e', 'r', 'a', "'", ' ', 't', 'ó', 'k', 'a', "'", ' ', 'n', 'ò', 'n', ':', 'w', 'a', ' ', 'e', 'n', 'w', 'á', 'h', 's',
'e', 'h', 't', 'e', "'", ' ', 'n', 'e', ' ', 'k', 'a', "'", 'n', 'i', 'k', 'o', 'n', 'h', 'r', 'a', 'k', 's', 'é', 'n', 'h', 't', 's', 'h', 'e', 'r', 'a', "'", '.']
contains item`

joanise:
### PR Goal? 

Fix the bug where a literal `/` in text would cause a crash, by replacing it with `_SLASH_` when the characters are joined with  `/`.

### Fixes? 

Fixes #460
Fixes #540

### Feedback sought? 

This `_SLASH_` is not just in internal state, it's also saved in `preprocessed/filelist.psv`, which means it's ultimately user visible. Because of that, I would like a strong agreement on what string to use here. Is `_SLASH_` OK or should I use `<SLASH>` or something else? I'm not worried about collisions, but whatever choice we make now will be permanent because changing it in the future will be a breaking change.

Beyond that, regular sanity checking.

### Priority? 

beta

### Tests added? 

yes

### How to test? 

Follow https://github.com/EveryVoiceTTS/EveryVoice/issues/460#issuecomment-2455849389
and see that `everyvoice preprocess` doesn't crash.

Or use `everyvoice/tests/data/metadata_slash_pipe.psv` in the wizard and see that `everyvoice preprocess` doesn't crash.

### Confidence? 

High for the coding, low for the choice of substitution string.

### Version change? 

no

### Related PRs? 

none


marctessier:
I ran some tests with some custom data  ,  this does fix the bug while preprocessing.  I  also did  a quick training run with success . No issues detected.

joanise:
Decision, in consultation with @SamuelLarkin : since we already use `<SIL>` in similar contexts, let's use `<SLASH>` to maintain better consistency.

SamuelLarkin (approved):
LGTM

PR #579, opened 2024-11-06
test: exercise finding wavs in nested symlink structures

### LINKED ISSUE #89: Add specific unit testing for symlinks in wavs directory.
#83 adds testing for the wavs directory in `test_wizard.py` method `test_wavs_dir()`, but it's not complete.

#67 flagged issues with globing over symlinks and we should add unit test cases in (or besides) `test_wavs_dir()` for when the wav files are symlinks and when the directory is a symlink, to make sure we don't cause #67 again.

joanise:
Fixes #89

### PR Goal? 

Unit test our glob code to make sure it works with deep symlink structures.

### Fixes? 

Fixed #89

### Feedback sought? 

rubber stamping

### Priority? 

beta

### Tests added? 

nothing but!

### How to test? 

If you change `WavsDirStep.validate()` to use `contains_wavs = next(path_expanded.glob("**/*.wav"), False)` instead of the current code, you'll see that `test_wizard` will fail on this new test function, but it passes with the current code.

```sh
python -m unittest \
  everyvoice.tests.test_preprocessing.PreprocessingTest.test_no_speaker \
  everyvoice.tests.test_wizard.WizardTest.test_wavs_dir
```

### Confidence? 

high

### Version change? 

no

### Related PRs? 

none

### DIFF
--- a/everyvoice/tests/test_preprocessing.py
+++ b/everyvoice/tests/test_preprocessing.py
@@ -494,7 +494,7 @@ class PreprocessingTest(PreprocessedAudioFixture, BasicTestCase):
         lj_preprocessed = tmpdir / "preprocessed"
         lj_filelist = lj_preprocessed / "preprocessed_filelist.psv"
 
-        fp_config = FeaturePredictionConfig(contact=self.contact)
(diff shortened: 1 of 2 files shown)


SamuelLarkin (approved):
LGTM

PR #582, opened 2024-11-12
feat: auto fixing from black

SamuelLarkin:
Will investigating #581 , I ran `pre-commit run --all` and `flake8` made some changes to some files

### PR Goal? 

Update some files changed by `flake8`.

### Feedback sought? 

merge approval

### Priority? 

low

### Tests added? 

None

### How to test? 

`pre-commit run --all`

### Confidence? 

good

### Version change? 

No

### Related PRs? 

depends on: https://github.com/EveryVoiceTTS/EveryVoice/pull/583

PR #585, opened 2024-11-13
fix: reconfigure flake8; it does not support pyproject.toml

joanise:
Refs:
 - https://github.com/PyCQA/flake8/issues/234
 - https://flake8.pycqa.org/en/latest/user/configuration.html

### PR Goal? 

The flake8 configuration done in `pyproject.toml` is ineffective, because flake8 just does not support that, so all the previously ignored warnings and the rest of our configuration are suddenly ignored.

This PR restores our previous flake8 config.

### Feedback sought? 

testing I guess

question: do we want this in `.flake8` or in `setup.cfg`? I didn't use the latter because it would be the only tool we configure in there.

### Priority? 

Before merging #582 I guess, so CI can pass for that PR.

### Tests added? 

n/a

### How to test? 

Running `pre-commit run --all` should show no warnings from flake8. It will show lots of issues from black (#582 will fix that) and mypy, please ignore those for the purpose of this PR which only touches flake8.

### Confidence? 

high

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,5 @@
+[flake8]
+ignore = E203, E266, E501, W503
+max-line-length = 88
+max-complexity = 18
+select = B, C, E, F, W, T4, B9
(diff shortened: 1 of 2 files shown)


SamuelLarkin (approved):
LGTM

PR #593, opened 2024-11-25
fix: exported model name should match generator class name

roedoejet:
### PR Goal? 

When you run `everyvoice export` the vocoder HiFiGAN is exported as a HiFiGANGenerator which then breaks the model name validation. 

### Fixes? 

I just fixed the bug before creating an issue report since I just noticed it this weekend.

### Feedback sought? 

@SamuelLarkin - I'm not sure if the quick fix I've done here is the best way of addressing this, can you please advise if there is another place I should address the fix?

### Priority? 

high - currently demos are broken with exported vocoders

### Tests added? 

I can't really test the exporting of a model because of the size of the vocoder model (1GB)

### Related PRs? 

https://github.com/EveryVoiceTTS/HiFiGAN_iSTFT_lightning/pull/42

### DIFF
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -37,6 +37,10 @@ from everyvoice.model.feature_prediction.FastSpeech2_lightning.fs2.cli.train imp
     train as train_fs2,
 )
 from everyvoice.model.vocoder.HiFiGAN_iSTFT_lightning.hfgl.cli import (
+    HFG_EXPORT_LONG_HELP,
+    HFG_EXPORT_SHORT_HELP,
+)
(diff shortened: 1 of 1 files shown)


joanise:
Samuel and I reviewed this together, and we agree it is implemented correctly, but there is some refactoring we want to apply before merging. PR coming soon for that.

joanise:
@roedoejet Please review https://github.com/EveryVoiceTTS/HiFiGAN_iSTFT_lightning/pull/43 and #595
Sam and I agreed this PR here was conceptually OK but needed refactoring before merging.

PR #596, opened 2024-11-26
Dev.ej/fix typer

joanise:
See https://github.com/EveryVoiceTTS/wav2vec2aligner/pull/16

Plus slow down the yaspin spinner update speed to reduce annoying flicker


SamuelLarkin (approved):
LGTM

PR #600, opened 2024-11-28
ci: simplify CI EV install, use apt-get, remove conda

joanise:
### PR Goal? 

We don't actually need conda, at least not anymore:
 - sox and ffmpeg can be installed with apt-get
 - previously optimized pip installs for pyworld and pycountry can simply be cached by the standard pip caching mechanism, instead of using custom optimizations

The result is a more standard CI workflow, and a tiny bit faster to boot.

### Feedback sought? 

Mostly rubber stamping

### Priority? 

normal

### Tests added? 

it's just CI

### How to test? 

look at CI workflows from this branch vs others

### Confidence? 

high

### Version change? 

no

### Related PRs? 

none

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,27 +12,17 @@ jobs:
     timeout-minutes: 15
     permissions:
       pull-requests: write
-    defaults:
-      run:
-        # Required for conda-incubator/setup-miniconda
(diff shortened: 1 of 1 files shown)


SamuelLarkin (approved):
LSTM

PR #602, opened 2024-11-28
docs: quickstart instructions for install from PyPI

joanise:
### PR Goal? 

On PyPI, like on GH, the quickstart instructions start with checking out EV from GitHub. We should start with PyPI-compatible instructions, then move on to instructions to install from source.

### Feedback sought? 

Make sure my updated instructions are correct

### Priority? 

medium

### Tests added? 

no, but I tested these instructions while writing #600

### How to test? 

Follow the new instructions and make sure the result is a valid environment.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

no

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,19 @@
 
 This is the Text-to-Speech (TTS) toolkit used by the Small Teams "Speech Generation for Indigenous Language Education" project.
 
-## Quickstart
+## Quickstart from PyPI
+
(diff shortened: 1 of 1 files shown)


joanise:
Reviewed with Marc and approved verbally.

This README is a bit longer than ideal, though, we should think about moving some parts of it out to the manual. Mind you, EV's readme is not the worst, g2p and Studio have more need for such docs refactoring.

PR #604, opened 2024-11-29
chore: update submodule to skip instead of fail tests on network failure

joanise:
See https://github.com/EveryVoiceTTS/wav2vec2aligner/pull/17


marctessier (approved):
Works as expexted.

With internet access: 
```
Ran 213 tests in 77.518s
OK
```

With no internet access:
```
Ran 213 tests in 64.817s

OK (skipped=1)
```

PR #612, opened 2024-12-19
Dev.ap/fix gradio

roedoejet:
### PR Goal? 

The demo is broken! This fixes it by passing the expected format of the wav samples to gradio. I also bumped gradio as discussed.

### Feedback sought? 

Test out the demo to make sure, but mostly sanity.

### Priority? 

high

### Confidence? 

high


joanise (approved):
This works, and let me figure out how to get the demo to save the .wav files

PR #613, opened 2024-12-19
ci: run licensecheck to auto-audit our dependencies

joanise:
Same as Studio and g2p, but with more exceptions whose licenses are not auto-detected.


roedoejet (approved):
lgtm!

joanise:
> it's failing on a bunch of nvidia deps though. not sure if you can add an ignore pattern

Yeah, I failed to notice these initially. I had to exclude "proprietary" globally, but since that's not necessarily safe I had to add custom logic using grep making sure we're not ignoring anything we haven't manually vetted.

Issue #611, opened 2024-12-19
Create a regression Test


joanise:
### Description & Motivation

Regression test overview:
 - Use LJ, maybe 10000 sentence sample
 - run new-project with a canned --resume-from  file
 - preprocess
 - train fs2
 - train a vocoder on same
 - synthesize a bunch of stuff
 - run the demo -- maybe use playwright to enter a couple sentences to synthesize

Synthesis variants to include:
 - synthesize with all output types
 - synthesize from text (characters, phonemes, phonological features)
 - synthesize from spectrograms
 - try synthesizing spectrograms and then running vocoder fine-tuning on the generated spectrograms

Issue #478, opened 2024-06-19
`everyvoice demo` creates folders in an active directory 


wiitt:
Running `everyvoice demo` creates logs_and_checkpoints and preprocessed/YourDataSet folders if they don't exist in a directory which had been active when the `everyvoice demo` command was activated.
Running speech synthesis in demo interface creates a synthesis_output folder with a wav folder and latest_predict_data.pth file in it.
Although demo doesn't save anything in the created folders, no folder creation should occur.

joanise:
It looks like `logs_and_checkpoints` no longer gets created, which is good, but the `syntesis_output` folder still gets created when you synthesize stuff, with this contents:
```
synthesis_output/
├── latest_predict_data.pth
└── wav/
```

joanise:
Here's another related bug, though: if you use -o, it's only half honoured:

Command: `everyvoice demo -o foo ../fs2.ckpt ../voc.ckpt`

File structure after synthesizing a file:
```
.
├── foo/
│   └── wav/
└── synthesis_output/
    └── latest_predict_data.pth
```

I think the desired behaviour is:
1) by default, no output is created
2) if -o is used, then maybe it's OK to create some output
3) since no files ever actually get saved into wav/, it should not be created
4) if it's relevant to create latest_predict_data.pth, it needs to be in the directory specified by -o, but I don't understand why that file gets saved, that probably should not happen, at least not by default.

roedoejet:
I don't know if I agree with "by default, no output is created" - since I believe the output is used for caching (although this assumption needs to be verified)

no files ever get saved into wav? I believe they get saved somewhere, no?

I also don't understand `latest_predict_data.pth` or why it's being saved unfortunately :(

marctessier:
This is still happening, extra empty folder    `logs_and_checkpoints` and also `preprocessed/YourDataSet` being created  that should not created or should  needed for the DEMO.


```
$ find logs_and_checkpoints/
logs_and_checkpoints/

$ find preprocessed/
preprocessed/
preprocessed/YourDataSet
```

PR #614, opened 2024-12-20
fix: let the demo save wav files to disk again

joanise:
### PR Goal? 

Working on https://github.com/EveryVoiceTTS/EveryVoice/issues/478 I noticed that the demo no longer outputs .wav files at all.

This PR restores just that functionality, minimally.

It doesn't address the question of the `.pth` file that gets generated, and that ignores the directory specified with `-o`.

### Feedback sought? 

I don't like this solution, because the wav writer callback gets created twice, one that gets called by the pytorch_lightning/trainer and another that gets instantiated and called directly in demo/app.

I would like to find a solution where
 - only one callback is created, and reused in the two places
 - even better, `synthesize_audio()` is only called once and its output is reused

But for the moment the question is, is this simple solution good enough for now?

### Priority? 

beta I guess, and whenever we want to have a demo that saves generated audio files.

### Tests added? 

no, unfortunately we have no demo tests. But #611 is my next priority and I hope I'll find a way to exercise the demo in that regression suite.

### How to test? 

Run the demo

    everyvoice demo fs2.cpkt voc.cpkt

and see the wav files generated in `synthesis_output/wav` again, or in `dir/wav` if you specify `-o dir`.

### Confidence? 

both high and low.

high: this PR gets .wav files generated

low: I don't like the solution

### Version change? 

no

### Related PRs? 

no

### DIFF
--- a/everyvoice/demo/app.py
+++ b/everyvoice/demo/app.py
@@ -20,6 +20,9 @@ from everyvoice.model.feature_prediction.FastSpeech2_lightning.fs2.model import 
 from everyvoice.model.feature_prediction.FastSpeech2_lightning.fs2.prediction_writing_callback import (
     PredictionWritingWavCallback,
 )
+from everyvoice.model.feature_prediction.FastSpeech2_lightning.fs2.type_definitions import (
+    SynthesizeOutputFormats,
+)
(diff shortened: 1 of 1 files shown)


joanise:
closing as Aidan is working on this via a different task

PR #615, opened 2024-12-20
perf: refactor wizard to avoid expensive imports at CLI startup

joanise:
### PR Goal? 

On the login node on GPSC-C, the CLI can take 4 or 5 seconds to respond to a `<tab>` because it's loading a lot of stuff. The biggest culprits are typer, rich, and other things required to actually provide CLI help and autocompletion, but I noticed the wizard is doing some imports we can defer too, so I refactored it.

This results in a 15% or so speed up of `everyvoice -h` once already loaded from cache, and I'm not sure how much on a first try when not in disk cache yet (that's harder to measure!). Not the big difference I was hoping for, but still worth committing and pushing.

Plus, the refactored code is a little cleaner than the original imho.

### Feedback sought? 

Regular review

The diff will look huge, but it's not: stuff removed from `wizard/__init__.py` has been moved verbatim to `wizard/tour.py`.

### Priority? 

low

### Tests added? 

was already covered, adjusted where needed

### How to test? 

run `everyvoice test cli` and maybe `everyvoice new-project` and see no failures.

### Confidence? 

high: this is just moving definitions from one file to another, using VSCode's refactoring tools and then testing and adjusting the few little bits it didn't handle.

Note: I thought of having the Step classes in `step.py` and the Tour class in `tour.py`, but doing so created circular imports since each uses the other for typing hints, and therefore I decided it made most sense to keep them all together in `tour.py`.

### Version change? 

no


roedoejet (approved):
Nice - thanks Eric! I agree it makes sense to move stuff out of the __init__ module. Moving things to tour.py makes sense to me. This PR looks good, and does seem to speed things up. Thanks!

PR #616, opened 2024-12-20
feat: write a regression test suite that can be sbatch submitted

joanise:
Starting to address #611

### PR Goal? 

Start to write a regression test suite that we can sbatch submit to the cluster with GPUs and whatever memory we need.

For now, I'm using a tiny dataset, but it's easy to adjust to something larger.

### Feedback sought? 

While this is early, you can already try to give it a spin: edit the two user config variables at the top if needed, and run 

    sbatch EveryVoice/everyvoice/tests/regression-test.sh

### Priority? 

beta I guess

### Tests added? 

nothing but

### Confidence? 

medium, this still needs more work

### Version change? 

no

### Related PRs? 

no

### TODO

 - [ ] use coverage analysis
 - [ ] add at least minimal checks to make sure the output files are as expected
 - [ ] run the demo and use playwright to click in it
 - [ ] add other variants -- there are lots of paths we can take through the EV pipeline!

### DIFF
--- a/everyvoice/tests/regression/go.sh
+++ b/everyvoice/tests/regression/go.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+#SBATCH --job-name=EV-r-main
+#SBATCH --partition=standard
+#SBATCH --account=nrc_ict
+#SBATCH --qos=low
+#SBATCH --time=10080
+#SBATCH --ntasks=1
+#SBATCH --cpus-per-task=1
+#SBATCH --mem=8000M
+#SBATCH --output=./%x.o%j
+#SBATCH --error=./%x.e%j
+
+# Automated application of the instructions in README.md
+
+set -o errexit
+
+TOP_LEVEL_DIR=$(mktemp --directory regress-$(date +'%Y%m%d')-XXX)
+cd "$TOP_LEVEL_DIR"
+
+../prep-datasets.sh
+for DIR in regress-*; do
+    pushd "$DIR"
+    sbatch ../../regression-test.sh
+    popd
+done
+
+coverage run -p -m everyvoice test
+
+JOB_COUNT=$(find . -maxdepth 1 -name regress-\* | wc -l)
+while true; do
+    DONE_COUNT=$(find . -maxdepth 2 -name DONE | wc -l)
+    if (( DONE_COUNT + 2 >= JOB_COUNT )); then
+        break
+    fi
+    echo "$DONE_COUNT/$JOB_COUNT regression job(s) done. Still waiting."
+    date
+    sleep $(( 60 * 5 ))
+done
+
+echo "$DONE_COUNT regression jobs done. Calculating coverage now, but some jobs may still be running."
+../combine-coverage.sh
+cat coverage.txt
+
+while true; do
+    DONE_COUNT=$(find . -maxdepth 2 -name DONE | wc -l)
+    if (( DONE_COUNT >= JOB_COUNT )); then
+        break
+    fi
+    echo "$DONE_COUNT/$JOB_COUNT regression job(s) done. Still waiting."
+    date
+    sleep $(( 60 * 5 ))
+done
+
+echo "All $DONE_COUNT regression jobs done. Calculating final coverage."
+rm .coverage
+../combine-coverage.sh
+cat coverage.txt
--- a/everyvoice/tests/regression/go.sh
+++ b/everyvoice/tests/regression/go.sh
@@ -18,10 +18,16 @@ set -o errexit
 TOP_LEVEL_DIR=$(mktemp --directory regress-$(date +'%Y%m%d')-XXX)
 cd "$TOP_LEVEL_DIR"
 
+if sbatch -h >& /dev/null; then
+    SUBMIT_COMMAND=sbatch
+else
+    SUBMIT_COMMAND=bash
+fi
+
 ../prep-datasets.sh
 for DIR in regress-*; do
     pushd "$DIR"
-    sbatch ../../regression-test.sh
+    $SUBMIT_COMMAND ../../regression-test.sh
     popd
 done
 
--- a/everyvoice/tests/regression/prep-datasets.sh
+++ b/everyvoice/tests/regression/prep-datasets.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+# Prepare the datasets and directories for our regression test cases
+
+set -o errexit
+
+# Usage: cat my_file | get_slice lines_to_keep > out
+# Use a number of lines or full to get all lines
+get_slice() {
+    lines=$1
+    if [[ $lines == full ]]; then
+        cat
+    else
+        head -"$lines"
+    fi
+}
+
+EVERYVOICE_REGRESS_ROOT=$(python -c 'import everyvoice; print(everyvoice.__path__[0])')/tests/regression
+
+LJ_SPEECH_DATASET=$HOME/sgile/data/LJSpeech-1.1
+LJSLICES="150 600 1600 full"
+for slice in $LJSLICES; do
+    dir=regress-lj-$slice
+    mkdir "$dir"
+    ln -s "$LJ_SPEECH_DATASET/wavs" "$dir"/
+    get_slice "$slice" < "$LJ_SPEECH_DATASET/metadata.csv" > "$dir"/metadata.csv
+    cp "$EVERYVOICE_REGRESS_ROOT"/wizard-resume-lj "$dir"/wizard-resume
+    cat <<'==EOF==' > "$dir"/test.txt
+This is a test.
+I am an anvil.
+I have no idea what to write here, but it has to be synthesizable text; so here is something!
+Boo!
+==EOF==
+    echo spec > "$dir"/test2.txt
+done
+
+SinhalaTTS=$HOME/sgile/data/SinhalaTTS
+dir=regress-si
+mkdir $dir
+ln -s "$SinhalaTTS/wavs" $dir/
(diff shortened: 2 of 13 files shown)


joanise:
TODO: add coverage analysis
TODO: add the demo and playwright to click in it
TODO: add other variants -- there are lots of paths we can take through the EV pipeline!

marctessier:
This test is hardcoded for GPSC7.   ( GPSC7 was busy so I modified for V100 , GPSC5) .

 I ran the test   

 I did get this message in my error_log , where I did not need... ( I can ignore this below..) 
```
cat EV-regress.e5091562
/var/lib/slurm-llnl/slurmd/job5091562/slurm_script: line 22: /home/tes001/start_ev.sh: No such file or directory
```


BUT I am getting this message:

`Your resume-from file is likely out of sync. Aborting.`

Can we add this to the script where the error and output log will be written where the script was ran using the jobname "EV-regress" .

```
#SBATCH --output=./%x.o%j
#SBATCH --error=./%x.e%j
```

`(EveryVoice_pr616) [U20-GPSC5]:$ cat EV-regress.o5091580
╭──────────────────────────────────────────────────────────────────────────────╮
│ Welcome to the EveryVoice Wizard. We will guide you through the process of   │
│ setting up the configuration for a new EveryVoice project.                   │
│                                                                              │
│ Navigation: as any point, you can hit Ctrl+C to: go back a step, view        │
│ progress, save progress, or exit the wizard.                                 │
│                                                                              │
│ From saved progress, you can resume at any time by running the same command  │
│ with the --resume-from option.                                               │
╰──────────────────────────────────────────────────────────────────────────────╯
Applying saved response 'regress' for Name Step
Great! Launching Configuration Wizard 🧙 for project named 'regress'.
Applying saved response 'EveryVoice Regressor' for Contact Name Step
Great! Nice to meet you, 'EveryVoice Regressor'.
Applying saved response 'ev@nrc.ca' for Contact Email Step
Great! Your contact email 'ev@nrc.ca' will be saved to your models.
Applying saved response '.' for Output Path Step
The Configuration Wizard 🧙 will put your files here: 'regress'
Applying saved response 'metadata.csv' for Filelist Step
Applying saved response 'Yes, I do have permission to use this data.' for 
Dataset Permission Step
Applying saved response 'psv' for Filelist Format Step
Error: next tour question is Filelist Text Representation Step but resume list 
has Filelist Has Header Line Step instead.
Your resume-from file is likely out of sync. Aborting.`

joanise:
Nice, thank you for this output, I'll take a look when I resume work on this.

joanise:
@marctessier I have fixed the issues you noticed, and more:
 - I was assuming the metadata file did not have a header, but you added one in your copy of LJ. I added code to skip the first line so my assumptions are met.
 - I now ignore the activate script if it's not found
 - I've reduce the tqdm logging frequency to every >=5s with `export TQDM_MININTERVAL=5`. Makes batch job logs nicer to read. (Still looking for the equivalent with torch, to make it show the loss on screen less often.)
 - I've merge stderr and stdout into a file in the current directory as you suggested.
 - I now log each command that gets run before it starts, so it's easier to follow the logs afterwards.

joanise:
Feedback from Samuel:
- [ ] fetching bundle.js from unpkg fails if the cluster proxy is not defined; needs graceful fall back like readalongs CLI
- [ ] make sure timing out trying to fetch from unpkg does mean a lot of time waiting for it
- [ ] LJ-150 fails to train with `everyvoice.utils:filter_dataset_based_on_target_text_representation_level:96 - Sorry you do not have enough characters data in your current validation filelist to run the model with a batch size of 16.`
- [ ] When training fails, exit right away, don't continue to do stuff that's guaranteed to fail
- [ ] in the readme.md, document that you can use `sbatch --partition=... --account=... ../../myscript.sh` to override the partition and account that are hard-coded in the scripts

everyvoice/tests/regression/go.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+#SBATCH --job-name=EV-r-main
+#SBATCH --partition=standard
+#SBATCH --account=nrc_ict
```
roedoejet, line 5: Maybe we can do what @SamuelLarkin does and provide the --account and --partition for all clusters.


```
# On GPSC5
#SBATCH --partition=standard
#SBATCH --account=nrc_ict
# On GPSC7
##SBATCH --partition=gpu_a100
##SBATCH --account=nrc_ict__gpu_a100
# On GPSC-C
##SBATCH --partition=gpu_v100
##SBATCH --account=nrc_ict__gpu_v100
# On Trixie
##SBATCH --partition=TrixieMain,JobTesting
##SBATCH --account=dt-mtp
```

roedoejet: (I just pulled this from one of the sample scripts that @SamuelLarkin sent - I'm not actually sure if these are the partitions we need)

everyvoice/tests/regression/prep-datasets.sh
```diff
@@ -0,0 +1,69 @@
...
+LJSLICES="160 600 1600 full"
+for slice in $LJSLICES; do
+    dir=regress-lj-$slice
+    mkdir "$dir"
+    ln -s "$LJ_SPEECH_DATASET/wavs" "$dir"/
+    get_slice "$slice" < "$LJ_SPEECH_DATASET/metadata.csv" > "$dir"/metadata.csv
+    cp "$EVERYVOICE_REGRESS_ROOT"/wizard-resume-lj "$dir"/wizard-resume
+    cat <<'==EOF==' > "$dir"/test.txt
+This is a test.
+I am an anvil.
```
roedoejet, line 32: 😆 ⚒️

everyvoice/tests/regression/regression-test.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+
+#SBATCH --job-name=EV-regress
+#SBATCH --partition=gpu_a100
```
roedoejet, line 4: same here with the partition/accounts for all clusters

everyvoice/tests/regression/regression-test.sh
```diff
@@ -0,0 +1,99 @@
...
+
+# Exercise two-step synthesis
+ONE_WORD=$(cat ../test2.txt)
+r "coverage run -p -m everyvoice synthesize from-text --output-type spec --text '$ONE_WORD' '$FS2'"
+r "coverage run -p -m everyvoice synthesize from-spec \
+    --input synthesis_output/synthesized_spec/'$ONE_WORD'-*.pt \
+    --model '$VOCODER'"
+
+# Exercise DeepForceAligner
+# Meh, this appears to be broken...
```
roedoejet, line 85: What's broken the aligner or coverage or? do you remember what the error is?

everyvoice/tests/regression/regression-test.sh
```diff
@@ -0,0 +1,99 @@
...
+#r "coverage run -p -m dfaligner train config/everyvoice-aligner.yaml --config-args training.max_steps=$MAX_STEPS --config-args training.max_epochs=$MAX_EPOCHS"
+#r "coverage run -p -m dfaligner extract-alignments"
+
+
+# 7: spin up the demo
+# everyvoice demo $FS2 $VOCODER &
```
roedoejet, line 91: this is a TODO I imagine? Maybe mark it with TODO inline?

`TODO:` has a special syntax highlighting with the Python extension, so I like to use that.

everyvoice/tests/regression/regression-test.sh
```diff
@@ -0,0 +1,99 @@
...
+
+
+# 8: use playwright to synthesize something using the demo
```
roedoejet, line 94: Same here: `TODO: 8: use playwright to synthesize something using the demo`

roedoejet (approved):
This looks great @joanise ! I would just say there are a couple small changes that would make it easier to run on different clusters, see my comments. thanks!

SamuelLarkin:
Not an issue but I seeing
```
Warning: saved progress file is for EveryVoice Wizard version '0.2.0a1', but
this is version '0.3.0'. Proceeding anyway, but be aware that the saved
responses may not be compatible.
```
when on the latest shouldn't version of EV, shouldn't all the versions be compatible?

everyvoice/tests/regression/regression-test.sh
joanise: When I tried to run that command, it failed, but I didn't investigate further yet. To reproduce, just take one of the LJ directories and run it, which is the thing to do eventually, since we do want to add dfa testing here.

joanise:
> when on the latest shouldn't version of EV, shouldn't all the versions be compatible?

Oh, that's a really hard target to try to impose. If we add a question to the wizard at some point, old resume files will all be broken. I didn't think of these files as artefacts we're supposed to maintain backwards compatibility for... That's why I issued the warning, though I suppose we could adjust the warnings to more specifically complain only if the version is older than the last time we made a breaking change in the wizard?

joanise:
But actually, maybe now that we've published, we can add our own requirement to auto-convert from published versions, and this regression test suite gives us the scaffolding to test for that.
My proposal, then: I'll find the oldest version of EV from which the resume files are still compatible, and adjust the warning not to say anything from that point forward, and let's set the new requirement that breaking changes to the wizard cannot break this regression suite.
That's going to make life a lot harder should we choose to add or remove a wizard question, but as we're approaching release it's a discipline we need to have.
And I guess if we make a breaking change to the wizard, we could write a filter function that adapts existing resume files, based on their declared version, with defaults for the new questions that make things work like before said breaking change.

PR #618, opened 2025-01-07
Dev.ej/fix docs ci

joanise:
### PR Goal? 

Deploy docs CI workflow was broken: see https://github.com/EveryVoiceTTS/EveryVoice/actions/runs/12655263006 which failed on the last push to main.

We updated all our other workflow to use setup-python instead of conda incubator, this PR does the same for the deploy docs workflows.

### Fixes? 

Broken docs deploy

### Feedback sought? 

sanity checking

### Priority? 

before the next PR merge to main, ideally

### Tests added? 

https://github.com/EveryVoiceTTS/EveryVoice/actions/runs/12656113424 ran with the docs deploy workflow, but with `--push` removed so it didn't push, and it completed successfully.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

no

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,25 +16,14 @@ jobs:
           token: ${{ secrets.SGILE_PAT }}
           submodules: recursive
           fetch-depth: 0 # fetch all commits/branches
-      - name: Set up Conda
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
Nice catch! I didn't notice. thanks @joanise !

PR #619, opened 2025-01-08
Refactor and fix the demo format code

joanise:
### PR Goal? 

Apply my suggestions to #617, make the code DRYer, fix the Spec download format, etc.

### Feedback sought? 

regular code review, testing

### Priority? 

same as #617

### Tests added? 

not yet, but as AP said that's work in progress coming soonish

### How to test? 

Spin up the demo and try to synthesize with every file format.

Read through at the output of `everyvoice demo -h`

Pass various combinations of valid and invalid values with -O add see what happens.

### Confidence? 

high

### Version change? 

This and #617 bring a new feature, so that would logically warrant a minor bump, but I think we've already got other such features. Soon I think?

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/106

### DIFF
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -616,19 +616,19 @@ def demo(
         ["all"],
         "--language",
         "-l",
-        help="Specify languages to be included in the demo. Example: everyvoice demo <path_to_text_to_spec_model> <path_to_spec_to_wav_model> --language eng --language fin",
+        help="Specify languages to be included in the demo. Must be supported by your model. Example: everyvoice demo T2S_MODEL VOCODER --language eng --language fin",
     ),
     speakers: List[str] = typer.Option(
         ["all"],
         "--speaker",
         "-s",
-        help="Specify speakers to be included in the demo. Example: everyvoice demo <path_to_text_to_spec_model> <path_to_spec_to_wav_model> --speaker speaker_1 --speaker Sue",
+        help="Specify speakers to be included in the demo. Must be supported by your model. Example: everyvoice demo T2S_MODEL VOCODER --speaker speaker_1 --speaker Sue",
     ),
     outputs: list[AllowedDemoOutputFormats] = typer.Option(
         ["all"],
         "--output-format",
         "-O",
-        help="Specify output formats to be included in the demo. Example: everyvoice demo <path_to_text_to_spec_model> <path_to_spec_to_wav_model> --output-format wav",
+        help="Specify output formats to be included in the demo. Example: everyvoice demo T2S_MODEL VOCODER --output-format wav --output-format readalong-html",
     ),
     output_dir: Path = typer.Option(
         "synthesis_output",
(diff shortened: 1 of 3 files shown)


everyvoice/cli.py
```diff
@@ -608,19 +616,19 @@ def demo(
         ["all"],
         "--language",
         "-l",
-        help="Specify languages to be included in the demo. Example: everyvoice demo <path_to_text_to_spec_model> <path_to_spec_to_wav_model> --language eng --language fin",
+        help="Specify languages to be included in the demo. Must be supported by your model. Example: everyvoice demo T2S_MODEL VOCODER --language eng --language fin",
```
roedoejet, line 619: I don't think I like `T2S_MODEL` since it's too close to TTS and that acronym doesn't appear anywhere else. I'd like to keep `text_to_spec` and `spec_to_wav` since it connects with the naming that we use in our config files and elsewhere in the CLI e.g.:

image

and `everyvoice train text-to-spec` for example.

You could change to `everyvoice demo [OPTIONS] TEXT_TO_SPEC_MODEL SPEC_TO_WAV_MODEL` if you don't like the angle brackets.

everyvoice/demo/app.py
```diff
@@ -358,10 +276,15 @@ def create_demo_app(
                 btn = gr.Button("Synthesize")
             with gr.Column():
                 out_audio = gr.Audio(format="wav")
-                out_file = gr.File(label="File Output")
+                if output_list == ["wav"]:
```
roedoejet, line 279: Maybe best to use `SynthesizeOutputFormats.wav` here instead of the `wav` string. I doubt this will ever change, but just in case.

roedoejet (commented):
I really like this! It's a solid re-factoring that makes the code cleaner and more understandable, and will make it easier to add additional methods in the future I think. I do disagree with the name change of the file arguments and left a comment for that. I also haven't tested it out and am leaving for the day soon, so I'll just comment on this PR for now and finish my review tomorrow. Thanks @joanise !

everyvoice/cli.py
joanise: Yeah, I hesitate for this one. I didn't like TEXT_TO_SPEC_MODEL  becoming `<path_to_text_to_spec_model>` as it did, but you're right, sticking to the name that occurs makes more sense. I'll fix this.

everyvoice/demo/app.py
joanise: I'll test if that works, if so sure, it's the same either way to me.
I'm not generally a fan of how long references to python enum values are, but it still makes sense to use them systematically when we define them.

marctessier:
This fixes the spec issue. Looks good. 

Also, I am no longer getting the  error in the server logs (  traceback...)  Nice work!



Eric and I had a hallway conversation about  the readalong-html file defaults being used...


Screenshot 2025-01-09 at 14 35 47


We are using the default "Title" -> `Your read-along title goes here`     and sub title =>  `Your read-along subtitle goes here`


We do have the name  / email address of the person who trained the model in the model file. We could use that information to help augment  the Tile / Subtitle info?

Issue #351, opened 2024-03-21
remind the user when the input wav has more than 2 channels


MENGZHEGENG:
As discussed with @roedoejet, when the input wav has more than 2 channels, an error message should be prompted to remind the user

roedoejet:
This code should go here to ensure that multi-channel audio causes an error that is readable to the user. 
https://github.com/EveryVoiceTTS/EveryVoice/blob/75cf5fee8ced7d978de62eed0521148c99e7fd59/everyvoice/preprocessor/preprocessor.py#L174

PR #626, opened 2025-01-20
feat: add channel check on the audio

MENGZHEGENG:
### PR Goal? 
Add check on the number of channels in the provided audio, remind the user when any audio has more than 2 channels (which can be problematic) and exit the program. Otherwise we will keep the only channel or average if there are two channels.

### Fixes? 
Fixs #351 

### Tests added? 
TODO

### DIFF
--- a/everyvoice/wizard/basic.py
+++ b/everyvoice/wizard/basic.py
@@ -1,4 +1,6 @@
 import json
+import sys
+import wave
 from pathlib import Path
 
 import questionary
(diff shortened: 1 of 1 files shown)


joanise:
Question about this test, counting channels: according to https://docs.python.org/3/library/wave.html, `wave` and `wave.open()` only support "uncompressed PCM encoded wave files".

@roedoejet Is EveryVoice already only compatible with such wav files, or could using `wave.open()` break some existing audio format flexibility?

I know that `wav2vec2aligner` can take .m4a audio files and align them, I've tried that before with success, but I don't know if they're supported anywhere else.

roedoejet:
> Question about this test, counting channels: according to https://docs.python.org/3/library/wave.html, `wave` and `wave.open()` only support "uncompressed PCM encoded wave files".
> 
> @roedoejet Is EveryVoice already only compatible with such wav files, or could using `wave.open()` break some existing audio format flexibility?
> 
I'm not 100% sure, but we use `torchaudio.load` elsewhere in EveryVoice to load and save audio, so I think we should use that here too

MENGZHEGENG:
There is another newer pull request on the same topic, so I close this one for now

Issue #357, opened 2024-03-26
Improve tone feature calculation for phonological features


roedoejet:
Currently the way we calculate tone features in `features.py` is very tied to Kanien'keha. We should generalized it.

Issue #329, opened 2024-03-13
We're not currently compatible with Python 3.12


joanise:
https://download.pytorch.org/whl/torch_stable.html does not include pre-compiled wheels for 2.1.0 with Python 3.12.
To support Python 3.12, we'll have to bump torch to >=2.2.0.

PR #627, opened 2025-01-20
Dev.ap/text pf updates expanded

roedoejet:
### PR Goal? 

This PR improves the phonological feature calculation by:

- increasing the number of features to include more punctuation (getting rid of BB and SB in favour of colon, semi-colon, and period)
- getting rid of custom tone feature calculation and just using Panphon's calculation using tone-bars. Note: this makes us require tone bars and not accents, but it also fixes https://github.com/EveryVoiceTTS/EveryVoice/issues/357
- we also add phonological feature calculation for common text encoding special tokens like [MASK] and [UNK]

### Fixes? 

https://github.com/EveryVoiceTTS/EveryVoice/issues/357

### Feedback sought? 

Sanity + high-level comments. low-level code comments are also welcome but less of a priority.

### Priority? 

medium

### Tests added? 

I've added some tests and updated existing ones

### How to test? 

A little bit tricky to test, but maybe inspecting some of the doctests will help. The main work is just analysing the updated code in text/features.py

### Confidence? 

medium (I think this is an improvement, and it is what we use in PF-BERT, but it's a big set of interrelated problems so I would appreciate the feedback)

### Version change? 

Yes, this is a breaking change.

### Related PRs? 

https://github.com/EveryVoiceTTS/DeepForcedAligner/pull/34
https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/109
https://github.com/EveryVoiceTTS/HiFiGAN_iSTFT_lightning/pull/52


roedoejet:
whoops - looks like some tests aren't passing - I'll investigate, but it should still be ready for review

roedoejet:
@joanise - I don't understand why the doctests are failling here, they're not failing on my machine, do you have any ideas?

joanise:
We didn't publish 0.3.0 yet, so this PR should update the 0.3 schemas, not create the 0.4 schemas.

everyvoice/text/features.py
```diff
@@ -1,69 +1,32 @@
...
-                tone_features.append([0, 0, 0, 0, 0, 0, 0])
-        return np.array(tone_features, dtype=np.float32)
+    def cls_token(self):
+        return self.get_features(["[CLS]"])[0]
+
+    def sep_token(self):
+        return self.get_features(["[SEP]"])[0]
+
+    def unk_token(self):
+        return self.get_features(["[UNK]"])[0]
```
joanise, line 29: These five functions aren't used anywhere in the code base. Why implement them?

everyvoice/text/features.py
```diff
@@ -75,38 +38,107 @@ def get_punctuation_features(self, tokens: list[str]) -> npt.NDArray[np.float32]
...
+                punctuation_features.append([0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0])
             elif char == self.punctuation_hash["exclamations"]:
-                punctuation_features.append([0, 0, 0, 0, 0, 0, 1, 0])
+                punctuation_features.append([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0])
             elif char in self.config.symbols.silence:
-                punctuation_features.append([0, 0, 0, 0, 0, 0, 0, 1])
+                punctuation_features.append([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1])
             else:
-                punctuation_features.append([0, 0, 0, 0, 0, 0, 0, 0])
+                punctuation_features.append([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
```
joanise, line 78: I'm really not a fan of this long if/elif list. Can we not construct a dict from punc hash value to these one-hot vectors with a loop, and do a dict look up for the punc entries, followed by the last elif and else?
I mean, this code works, so you don't have to change it, but it would be nicer to refactor it.

pyproject.toml
```diff
@@ -19,7 +19,7 @@ maintainers = [
   { name = "Samuel Larkin", email = "Samuel.Larkin@nrc-cnrc.gc.ca" },
 ]
 readme = "README.md"
-requires-python = ">=3.10, <3.12"
+requires-python = ">=3.10, <3.13"
```
joanise, line 22: This is a major change, it should not be hidden in an unrelated PR like this. It's fantastic news if we're now compatible with 3.12, but that should be in its own PR flagging it as a new feature, which fixes #329, and adding 3.12 to the classifiers in pyproject.toml.

joanise: And now that we're compatible with three different Python versions, we should add matrix testing to keep them all exercised, which is another good reason to splice it out of this PR.

joanise (commented):
sorry, a bunch of comments suggesting changes.

everyvoice/text/features.py
roedoejet: They're used in PF-BERT and possibly with other BERT-style text encoding

everyvoice/text/features.py
roedoejet: Hm, I kind of like just being able to see the full embedding like this - let's talk at the meeting today

joanise (commented):
I haven't been able to test yet, just tiny comments.

joanise (changes requested):
Since this PR breaks fs2 and hfgl models, it will need to include targeted messages when a user tries to load a model created with an older version of EV that's not compatible with the current punctuation lists.

joanise (commented):
I think this generally looks good, we're getting close to be able to merge. Just a couple small comments in the related PRs.

PR #629, opened 2025-01-30
feat: adjust the wizard resume message for known compatibility

joanise:
### PR Goal? 

The warning about the resume file coming from a different version is too generic. We know 0.2.0 and 0.3.0 resume files are compatible, so adjust the warning to pay attention the which versions are compatible and not.

### Feedback sought? 

Are the two new messages clearer, and less likely to cause confusion?

### Priority? 

beta

### Tests added? 

yes

### How to test? 

Take an existing save wizard progress file, and edit line 2 to show version 0.3.0 (current), 0.2.0 (older and still compatible) or 0.1.0 (older, not compatible) and see no message, still compat message, or not compat message, respectively.

### Confidence? 

medium

### Version change? 

no

### Related PRs? 

none

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,7 +99,6 @@ jobs:
 
   matrix:
     strategy:
-      fail-fast: false
       matrix:
         python-version: ["3.11", "3.12"]
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


SamuelLarkin (approved):
LGTM

PR #631, opened 2025-02-01
fix(demo): the demo was broken due to a mismatch between outputs and …

roedoejet:
The demo was broken, this is a quick friday-night fix for it, but I'll need fresh eyes on Monday. priority is high because it's blocking the demo transfer to WSB.

### DIFF
--- a/everyvoice/demo/app.py
+++ b/everyvoice/demo/app.py
@@ -46,6 +46,7 @@ def synthesize_audio(
     allowlist,
     denylist,
     output_dir=None,
+    include_file_output=True,
 ):
     if text == "":
         raise gr.Error(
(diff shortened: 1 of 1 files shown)


joanise:
The code looks good, but I don't understand what was broken. Can you describe the bug this fixes?

joanise (approved):
Tested, and ready to approve based on testing results.
 - I was able to get the demo to give me errors on `origin/main`
 - The same error did not occur with this PR
 - The code looks reasonable

roedoejet:
> The code looks good, but I don't understand what was broken. Can you describe the bug this fixes?
Yea, so there were two things - one, we always had the dropdown box for file format, even when wav was the only option. This does away with that, but then also has to optionally return one result from `synthesize_audio` instead of two since the file format output box is gone.

The other, more important issue is that for some reason we had `outputs=[out_audio],` which would NEVER return the file, it would only return the output wav. By changing this to `outputs=outputs` we return a list that has either just the output audio or the output audio AND the other file output.

joanise:
> The other, more important issue is that for some reason we had `outputs=[out_audio],` which would NEVER return the file, it would only return the output wav. By changing this to `outputs=outputs` we return a list that has either just the output audio or the output audio AND the other file output.

What is strange to me is that the demo used to work, with multiple download formats: we tested it when you added the readalongs to the demo. So when and how did it break?

But no matter, it's fixed again, that's the important.

PR #640, opened 2025-04-23
mention schemas for vscode before delving into vim, and modernize how docs deps are installed

joanise:
### PR Goal? 

Several related things:
 - https://docs.everyvoice.ca/stable/guides/schemas/ is a bit confusing, because we don't talk about schemas in general, we just jump into installing them in vim. Let's have an intro to this page.
 - To test the above, I naturally did `pip install -e .[docs]`, as it should be, but it didn't work. Now it does.
 - I find it unfriendly when navigation is only possible from the TOC, so I added to footer with previous/next links.

### Feedback sought? 

normal review

### Priority? 

low

### Tests added? 

n/a

### How to test? 

```
pip install -e .[docs]
mkdocs serve
```

and then review the update page at http://127.0.0.1:8000/guides/schemas/

### Confidence? 

high

### Version change? 

no

### Related PRs? 

no

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install dependencies and package
         run: |
           CUDA_TAG=cpu pip install -r requirements.torch.txt --find-links https://download.pytorch.org/whl/torch_stable.html
-          pip install -e . -r docs/requirements.txt
+          pip install -e .[docs]
       - name: Setup doc deploy
         run: |
           git config user.name 'github-actions[bot]'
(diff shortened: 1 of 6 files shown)


joanise:
Oh, and I didn't realize g2pcallable was new to this PR. I thought the mypy errors were due to old typing issues, which is why I fixed it, but I see it's just WIP stuff at the moment,

SamuelLarkin (approved):
LGTM
Thanks for updating `pyproject.toml` with `[docs]` and dropping the `requirements.txt`. 😃

PR #643, opened 2025-04-25
perf: install pyworld-prebuilt for speed and robustness

joanise:
pyworld-prebuilt uses CI to build all the pre-compiled wheels for pyworld and therefore it is faster and more reliable to install, providing the same access to World.

### PR Goal? 

1) Replace `pyworld` by the `pyworld-prebuilt` fork which builds precompiled wheels in CI and provides them on PyPI.

Benefits:
 - much faster install: on GPSC, `pip install pyworld-prebuilt` just took 2.7s whereas `pip install pyworld` failed with a compilation error after 1m40s. When rerun in a better configured session, it took 30s just to compile the wheel, on top of the download.
 - `pip install pyworld` fails under various circumstances -- e.g., on GPSC in the example quoted above -- because compilation has a number of dependencies that are apparently not carefully declared by the package's build configuration.
 - works with `uv` (`uv pip install pyworld` always fails for me, `uv pip install pyworld-prebuilt` succeeds)

2) Simplify `make-everyvoice-env` to strip stuff we no longer need
 - lit: long since not needed
 - cython: I think it hasn't been needed for a while
 - pyworld special handling: no longer needed with pyworld-prebuilt

3) Update the installation instructions with shorter versions that are now both possible and quite reliable. And generally revamp them, too.

### Fixes? 

Flakiness issues with pyworld installation.

### Priority? 

normal

### Tests added? 

Our use of pyworld in preprocessor.py is already exercised in unit testing, confirming that the drop-in replacement works.

### How to test? 

```
pip uninstall pyworld
pip install -e .            # see pyworld-prebuilt installed
python -m unittest everyvoice.tests.test_preprocessing.PreprocessingTest
```

### Confidence? 

high

### Version change? 

no

### Related PRs? 

no

PR #647, opened 2025-05-01
Fix `--diagnostic` to work with `uv`

SamuelLarkin:
We moved to `uv` in #643 thus I created a `conda` free shell which fails with `everyvoice --diagnostic`.

### PR Goal? 

#643 

### Feedback sought? 

Code review

### Priority? 

low

### Tests added? 

No

### How to test? 

```sh
everyvoice --diagnostic
```
or
```sh
python -m unittest everyvoice.tests.test_cli.CLITest.test_diagnostic
```

### Confidence? 

fair

### Version change? 

No

### Related PRs? 

None

### DIFF
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -116,6 +116,53 @@ app = typer.Typer(
 )
 
 
+def _diagnostic_pip(use_uv: bool) -> bool:
+    """
+    Try to list installed packages using `uv` or `pip`.
+    """
+    try:
+        result = subprocess.run(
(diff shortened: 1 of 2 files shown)


joanise (commented):
Thanks, this is very helpful.

For unit testing, we should run `everyvoice --diagnostics`, like I think I already do, but adjust the assertion so it's happy with any package list output. I don't feel the need to CI-test all three ways to install packages just to test this.

joanise (approved):
Looks good, thanks!

PR #648, opened 2025-05-02
Dev.ej/regress rebooted

joanise:
### PR Goal? 

The overall goal is to improve the regression test suite in various ways.

1) improve the coverage reports
2) use playwright to exercise the demo in testing
3) factor out how to submit regression on different clusters
4) add a mix-corpus test case in the regression suite

I know it would be better to splice this is 2 or 3 PRs, but I didn't leave myself enough time to do that before Jordan arrives and I'm hoping he'll work on top of this branch, so I'm submitting it now like this.

### Feedback sought? 

Flag egregious issues only. This is still work in progress anyway, but I think it's in a state that's already worth merging, and we'll keep improving it in future PRs.

### Priority? 

high

### Tests added? 

it's all just testing

### How to test? 

run the regression suites on a cluster

### Confidence? 

high that it's ok to merge now, but I know we'll keep improving on this.

### Version change? 

no

### Related PRs? 

no

### DIFF
--- a/everyvoice/tests/regression/test-demo-app-lj-full.py
+++ b/everyvoice/tests/regression/test-demo-app-lj-full.py
@@ -1,0 +1,53 @@
+"""
+Test the Demo App with the regress-lj-full demo model with a denylist.
+To run this test:
+ - Run the regression test suite to get the regress-lj-full model.
+ - cd regression/regress-<date>-<suffix>/regress-lj-full/regress
+ - everyvoice demo logs_and_checkpoints/FeaturePredictionExperiment/base/checkpoints/last.ckpt logs_and_checkpoints/VocoderExperiment/base/checkpoints/last.ckpt --denylist <(echo test) --output-format wav
+ - python test_demo_app_mix.py
+"""
+
+from unittest import TestCase, main
+
+from playwright.sync_api import expect, sync_playwright
+
+
+class TestDemoAppErros(TestCase):
+    def test_rundemo(self) -> None:
+        with sync_playwright() as playwright:
+            browser = playwright.chromium.launch(headless=True)
+            context = browser.new_context()
+            page = context.new_page()
+            page.goto("http://127.0.0.1:7860/")
+            synthesize_button = page.get_by_role("button", name="Synthesize")
+            synthesize_button.click()
+            expect(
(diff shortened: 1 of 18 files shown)


everyvoice/tests/regression/format-coverage.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
+
+"""
+For some reason, the coverage reports in regression tend to include the full paths.
+This script filters them to just start with everyvoice/... and realigns the header
+and footer accordingly.
+
+Usage: coverage report --include='*/everyvoice/*' | python format-coverage.py
```
roedoejet, line 8: I think we could do this natively within coverage by putting:

```
[run]
relative_files = true
```

in our .coveragerc config. See https://github.com/nedbat/coveragepy/issues/948

everyvoice/tests/regression/test-demo-app-lj-full.py
```diff
@@ -0,0 +1,53 @@
+"""
+Test the Demo App with the regress-lj-full demo model as some exclusion options.
```
roedoejet, line 2: I'm not sure I understand this sentence "as some exclusion options" ?

roedoejet (approved):
Code looks great. Just a couple of minor comments. I haven't tested this as it sounds like @marctessier is. Thanks Eric - really nice to see playwright tests up for this!

everyvoice/tests/regression/format-coverage.py
joanise: This was a while back, but I'm pretty sure I tried this and it failed for me. But I should test this again before merging.

everyvoice/tests/regression/test-demo-app-lj-full.py
joanise: Neither do I! I'll have to try to remember what I meant here... Must be a typo.

marctessier:
This looks good now,  No more errors!  Nice work!

```
----------------------------------------------------------------------
Ran 231 tests in 95.540s

OK
============ Finished job 5234311 on Thu May 29 10:21:11 AM EDT 2025 with rc=0
```

PR #661, opened 2025-05-27
feat: support uv in make-everyvoice-env

joanise:
### PR Goal? 

Support `uv` in make-everyvoice-env

I also changed the default Python version to 3.12 (but I wonder if I should splice that out into at least a separate commit).

### Feedback sought? 

Test and make sure it works

### Priority? 

low

### Tests added? 

We don't have a test framework for this script, unfortunately.

### How to test? 

Run `./make-everyvoice-env --uv` with various option combinations.

Logic of note:
 - if uv, sox and ffmpeg are found, uses uv by default, otherwise conda
 - when using uv, I explicitly test for libsox.so via importing sox_effects and error out if it's not found
 - the default venv when using uv is `./.venv` because that's uv's default, and VSCode auto-detects it, which is convenient.

### Confidence? 

pretty high

### Version change? 

no

### Related PRs? 

none

### DIFF
--- a/everyvoice/run_tests.py
+++ b/everyvoice/run_tests.py
@@ -69,6 +69,7 @@ def all_test_suites() -> TestSuite:
     for submodule_testsuite in SUBMODULE_SUITES.values():
         suite = loader.discover(
             os.path.dirname(__file__) + submodule_testsuite[0],
+            top_level_dir=os.path.dirname(__file__),  # MANDATORY
         )
         test_suite.addTests(suite)
 
--- a/everyvoice/tests/test_text.py
+++ b/everyvoice/tests/test_text.py
@@ -19,6 +19,7 @@ from everyvoice.text.phonemizer import (
     DEFAULT_G2P,
     CachingG2PEngine,
     get_g2p_engine,
+    make_default_g2p_engines,
 )
 from everyvoice.text.text_processor import JOINER_SUBSTITUTION, TextProcessor
 from everyvoice.utils import (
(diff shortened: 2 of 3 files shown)


marctessier:
I am just starting testing:




```
./make-everyvoice-env --help

Usage: ./make-everyvoice-env [options]

  Create an environment for EveryVoice, automating all the instructions in
  README.md

Options:
  -h, --help              Print this help message

Torch pre-compilation options:
  --cuda CUDA_VERSION     Install torch compiled for CUDA_VERSION
                          Default: --cuda 11.8
  --cpu                   Install torch for use on CPU only
  --python PYTHON_VERSION Specify the Python version to use
                          Default: --python 3.12

Target Environment Specification:
  -n, --name ENV_NAME     Name of the conda environment to create
                          Default with --conda: --name EveryVoice
  -p, --prefix, --path ENV_PATH   Path of the environment to create
                          Default with --uv: --path ./.venv

Venv manager options:
  --uv                    Use uv to create a virtual environment
                          Default if uv, sox and ffmpeg are found in PATH.
  --conda                 Use conda to create a conda environment
                          Default if any of uv, sox or ffmpeg are not found.

```


Is it possible that the  `"Venv manager options:"  part ` is messing the short hand switch ( single minus - )     I see only the `--uv `and `--conda`   where all the other switches includes the short hand to keep consistant?


` --help` says  uv is default...  I should not need to specify according to the help... `--uv` or `--conda`   

```
(EveryVoice_pr661_uv) $ ./make-everyvoice-env -n EV_none
ERROR: --name is not compatible with --uv.
Use -h for help.
```

marctessier:
I was playing a bit dirty with the test above where I was using a conda environment where only sox and uv was installed.  I deactivated that ENV and tried again to see:

```
(base) $ ./make-everyvoice-env -n EV_none
Please make sure the CUDA version installed on your system matches CUDA_VERSION=11.8.
Creating EveryVoice environment called "EV_none" for CUDA 11.8 using Python 3.12 with conda.
Proceed (y/[n])? y
Proceeding
./make-everyvoice-env: line 212: syntax error near unexpected token `&
```

marctessier:
Also does not like it when I specify  `--conda`


```
(base) $ ./make-everyvoice-env --conda -n EV_conda
Please make sure the CUDA version installed on your system matches CUDA_VERSION=11.8.
Creating EveryVoice environment called "EV_conda" for CUDA 11.8 using Python 3.12 with conda.
Proceed (y/[n])? y
Proceeding
./make-everyvoice-env: line 212: syntax error near unexpected token `&'
```

Issue #655, opened 2025-05-26
Use subsample to enhance the regression test suite


joanise:
With #649 done, use the subsample script in regression to enhance how we create the test cases.

See `prep-datasets.sh`

1) for LJ, we currently take 4 slices based on a number of lines. Instead, take slices of 15 minutes, 60 minutes, and 180 minutes, as well as the full set, keeping the rest of the logic as it is.

2) for isiXhosa and Sinhala, add slices like we do for LJ, also 15, 60, 180 minutes, and full.

At the moment, this work can only be done on the `dev.ej/regress-rebooted` branch, but with `dev.jl/subsample` merged into it, so  the work should be started on such a branch, at least until those two branches are merged.

Note: the `regress-mix` case does not need to be modified.

PR #654, opened 2025-05-23
Dev.jl/subsample

jordanxlau:
### PR Goal? 
A new script was written in tests/regression called subsample.py. This script, intended to be used via the command line, will subsample

### Fixes? 
Issue #649 

### Feedback sought? 
Any feedback is appreciated. I'm not sure if I have covered all possible error cases.

### Priority? 
medium

### Tests added? 
See test_subsample.py.

### How to test? 

This is a standalone script. Test it by running commands such as:
`python subsample.py metadata.psv /wavs --duration 7 --format psv --header`

Try different formats such as tsv, csv, festival, with and without header, different durations, etc.

### Confidence? 
medium

### Version change? 
This is a minor change

### Related PRs? 
N/A

PR #662, opened 2025-05-28
Dev.jl/regress subsample

jordanxlau:
### PR Goal? 

Regression test script prep-datasets.sh was edited to produce regression testing datasets:
regress-\<dataset\>-\<duration\> subsets at 15 mins, 60 mins, 180 mins and full dataset

### Fixes? 

Issue #655

### Feedback sought? 

Please run the regression tests and confirm they work properly/pass.

### Priority? 

medium

### Tests added? 

N/A no python code in this PR

### How to test? 

Run the regression tests. Manually check the new datasets created.

### Confidence? 

This code cannot be merged without some git history revision due to it being reliant on #648 and #654 

### Version change? 

Minor

### Related PRs? 

PR #648 
PR #654

### DIFF
--- a/everyvoice/tests/regression/prep-datasets.sh
+++ b/everyvoice/tests/regression/prep-datasets.sh
@@ -4,28 +4,24 @@
 
 set -o errexit
 
-# Usage: cat my_file | get_slice lines_to_keep > out
-# Use a number of lines or full to get all lines
-get_slice() {
-    lines=$1
-    if [[ $lines == full ]]; then
-        cat
-    else
-        head -"$lines"
-    fi
-}
-
 EVERYVOICE_REGRESS_ROOT=$(python -c 'import everyvoice; print(everyvoice.__path__[0])')/tests/regression
 
 SGILE_DATASET_ROOT=${SGILE_DATASET_ROOT:-$HOME/sgile/data}
 
+DURATIONS="20 60 180 full"
+
 LJ_SPEECH_DATASET=$SGILE_DATASET_ROOT/LJSpeech-1.1
-LJSLICES="160 600 1600 full"
-for slice in $LJSLICES; do
-    dir=regress-lj-$slice
+
+for duration in $DURATIONS; do
+    dir=regress-lj-$duration
     mkdir "$dir"
     ln -s "$LJ_SPEECH_DATASET/wavs" "$dir"/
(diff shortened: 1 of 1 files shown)


everyvoice/tests/regression/prep-datasets.sh
```diff
@@ -55,13 +57,20 @@ cat <<'==EOF==' > "$dir"/test.txt
 තතො නං දුක්‌ඛමන්‌වෙති, චක්‌කංව වහතො පදං.
 ==EOF==
 echo "අක-ෂර" > "$dir"/test2.txt
+done
```
joanise, line 60: Except for the four lines of text and the line with just `==EOF==`, the code above this `done` line should be indented to match the loop it's in. It took me a moment to figure out why there was a `done` here with the lines above not indented I didn't immediately see it was ending the loop.

everyvoice/tests/regression/prep-datasets.sh
```diff
@@ -70,6 +79,7 @@ uqeqesho
 iimpumlo
 ==EOF==
 echo isiXhosa > "$dir"/test2.txt
+done
```
joanise, line 82: Same here, the literal text going into test.txt, and the `==EOF==` have to sit against the left margin, but not the bash code.

everyvoice/tests/regression/prep-datasets.sh
```diff
@@ -4,28 +4,24 @@
...
-    else
-        head -"$lines"
-    fi
-}
-
 EVERYVOICE_REGRESS_ROOT=$(python -c 'import everyvoice; print(everyvoice.__path__[0])')/tests/regression
 
 SGILE_DATASET_ROOT=${SGILE_DATASET_ROOT:-$HOME/sgile/data}
 
+DURATIONS="15 60 180 full"
```
joanise, line 11: Turns out we'll have to use 20, 60, 180 and full for now.

joanise (changes requested):
This is good, thanks! A few comments below.

Last changes to do:
 - rebase it over the most recent changes in main
 - there's a bug in my code in this whole regression suite, in how I let the default validation set selection happen. Aidan created issue #664 about it, but in the meantime for this PR, please change the shortest slice to 20 minutes instead of 15.

If you look at the output when you run the regression suites, you might notice (if you dig deep) that the lj 15 and sinhalla 15 cases are failing quite early. Using 20 will do as a temporary solution until we do the much longer work required for #664.

PR #670, opened 2025-06-11
docs: added EveryVoice icon

jordanxlau:
### PR Goal? 

I've added a logo to the EveryVoice docs.

### Fixes? 

Not currently associated with any issue.

### Feedback sought? 

Any

### Priority? 

low

### Tests added? 

N/A

### How to test? 

Open the EveryVoice docs locally in a few browsers.

### Confidence? 

medium

### Version change? 

no version change necessary

### Related PRs? 

None


jordanxlau:
image

roedoejet (changes requested):
Cool! nice idea @jordanxlau - since this is an svg though, a higher resolution option would be to copy the svg element:

https://app.screencast.com/jL5dtx1lug8ek

You could then paste the contents into a file and call it `logo.svg` or something and convert from there:

logo

Additionally, rather than editing the head directly, I think we should just edit the mkdocs.yml like so: https://squidfunk.github.io/mkdocs-material/setup/changing-the-logo-and-icons/?h=favicon#favicon

PR #671, opened 2025-06-11
Dev.ej/misc

joanise:
### PR Goal? 

remove the broken favicon setting

provide install instructions that work on Mac with zsh (the default shell): `pip install -e .[dev]` does not work on zsh, you need to quote it like `pip install -e '.[dev]'`.


roedoejet (approved):
lgtm. thanks @jordanxlau for putting us down this path!

Issue #638, opened 2025-04-15
Synthesize long-form audio by improving text processing


roedoejet:
### Description & Motivation

We should improve our handling of long-form synthesis by including something similar to https://github.com/fakerybakery/txtsplit

Below I have an example that includes commas as lower-priority fallbacks. It would be nice to integrate this with our TextConfig in a better way.

```python
import re

def txtsplit(text, desired_length=100, max_length=200):
    """
    Split text into chunks of approximately `desired_length`,
    trying to preserve sentence boundaries (!?.), with commas as lower-priority fallbacks.
    """
    # Normalize input
    text = re.sub(r'\n\n+', '\n', text)
    text = re.sub(r'\s+', ' ', text)
    text = re.sub(r'[“”]', '"', text)

    chunks = []
    current = ""
    pos = 0
    end = len(text)
    in_quote = False
    strong_splits = []
    weak_splits = []

    def peek(offset):
        idx = pos + offset
        if 0 <= idx < end:
            return text[idx]
        return ''

    def add_split(type_):
        if type_ == 'strong':
            strong_splits.append(len(current))
        elif type_ == 'weak':
            weak_splits.append(len(current))

    def commit():
        nonlocal current, strong_splits, weak_splits
        if current.strip():
            chunks.append(current.strip())
        current = ""
        strong_splits = []
        weak_splits = []

    while pos < end:
        char = text[pos]
        current += char

        # Handle quote toggling
        if char == '"':
            in_quote = not in_quote

        # Record potential split points
        if not in_quote:
            if char in '!?\n':
                add_split('strong')
            elif char == '.' and peek(1) in '\n ':
                add_split('strong')
            elif char in ',;:':
                add_split('weak')

        # If chunk is too long, find best place to split
        if len(current) >= max_length:
            split_at = None
            if strong_splits and len(current) > desired_length / 2:
                split_at = strong_splits[-1]
            elif weak_splits and len(current) > desired_length / 2:
                split_at = weak_splits[-1]

            if split_at:
                chunks.append(current[:split_at].strip())
                current = current[split_at:].lstrip()
            else:
                # fallback: force split at max_length
                chunks.append(current.strip())
                current = ""

            strong_splits = []
            weak_splits = []

        # Early commit on strong boundary
        elif not in_quote and (
            char in '!?\n' or (char == '.' and peek(1) in '\n ')
        ):
            if len(current) >= desired_length:
                commit()

        pos += 1

    if current.strip():
[...]

PR #672, opened 2025-06-12
[WIP] dev.jl/text chunking

### LINKED ISSUE #658: Integrate the splitter into EveryVoice synthesis
### Description & Motivation

This task is to optionally (we should be able to turn this off) split text into utterances of a target length.

We agreed this function should go in `everyvoice/text/splitter.py` and the licensing should reflect that it was adapted from txtsplit.

This should result in synthesis on both the command line, and the demo gradio app being synthesized in chunks. This will be easier in the demo since there's always only a batch size of 1 there, but it will be harder in the command line since we need to batch-synthesize.

The pseudo-code for the B=1 case is:

```python
audio_segments = []
for t in split_text(raw_text):
    audio_segments.append(model(t))
concatenated_audio = np.concatenate(audios)
```

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### LINKED ISSUE #659: Integrate the text splitting with TextConfig
### Description & Motivation

The text splitter works by determining weak and hard boundaries. These weak and hard boundaries are currently hard-coded into the function, but we should be able to define them in the text configuration, and we should infer reasonable defaults from the configuration as well. I.e. we need to replace the assumption that ? and ! are the only possible strong boundaries and :;, are the only possible weak boundaries with information from the text config.

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

jordanxlau:
### PR Goal? 

Integrate a text splitter that splits text inputs into smaller chunks before passing them to the neural network. The aim is to improve EveryVoice's synthesis of longer texts. It is not currently optional, that will be a later commit to this PR.

### Fixes? 

Fixes #658 
Will eventually fix #659 
Parent issue: #638 

### Feedback sought? 

Is the code clean, logical and easy to read?

### Priority? 

medium

### Tests added? 

Yes, `class TestTextSplit` in `test_text.py` was added.

### How to test? 

Synthesize long utterances (multiple batches) in the demo and via the command line.

### Confidence? 

medium

### Version change? 

Currently, text chunking cannot be turned off, and is a breaking change.

### Related PRs? 

There is a corresponding FastSpeech2_lightning PR, https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/113

### DIFF
--- a/everyvoice/demo/app.py
+++ b/everyvoice/demo/app.py
@@ -91,12 +91,12 @@ def synthesize_audio(
         filelist=None,
         filelist_data=None,
         teacher_forcing_directory=None,
-        batch_size=1,
+        batch_size=16,
         num_workers=1,
     )
 
     wav_writer = callbacks[SynthesizeOutputFormats.wav]
-    wav_output = wav_writer.get_filename(basename, speaker, language)
+    wav_output = wav_writer.last_file_written
 
     file_output = None
     if output_format != SynthesizeOutputFormats.wav:
--- a/everyvoice/tests/test_text.py
+++ b/everyvoice/tests/test_text.py
@@ -22,6 +22,7 @@ from everyvoice.text.phonemizer import (
     make_default_g2p_engines,
 )
 from everyvoice.text.text_processor import JOINER_SUBSTITUTION, TextProcessor
+from everyvoice.text.textsplit import chunk_text
 from everyvoice.utils import (
     collapse_whitespace,
     generic_psv_filelist_reader,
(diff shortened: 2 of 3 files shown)


marctessier (approved):
I only did some simple testing using LJ  and everything worked. 

I like in the logs how we can see how it's being "chunked" .   In this example it did it in groups of "3" and I can also hear the pattern being repeated in the audio produced.  I also tried something longer / using an other real data source .  I also find it produced a smoother more consistent  synth  the more longer the test when I compared :-)  

first example: 
`Processing text: ['This is a test to see how this will work. This is a test to see how this will work. This is a test to see how this will work.', 'This is a test to see how this will work. This is a test to see how this will work. This is a test to see how this will work.', 'This is a test to see how this will work. This is a test to see how this will work. This is a test to see how this will work.', 'This is a test to see how this will work. This is a test to see how this will work. This is a test to see how this will work.', 'This is a test to see how this will work. This is a test to see how this will work. This is a test to see how this will work.', 'This is a test to see how this will work. This is a test to see how this will work. This is a test to see how this will work.', 'This is a test to see how this will work. This is a test to see how this will work.']`

I also ran this on my laptop where less Vram is available with no issues... ( I noticed we changed the default batch size)  . ( wondering  if we should also changed num_workers=1 to 4     but  more test would be needed to see if  might  help speed up a little if your gpu can take it ?)   When on cpu it will by default try to use everything available...

PR #674, opened 2025-06-12
docs: update schemas and version to 0.2.0

roedoejet:
We actually hadn't published a non-alpha release of 0.2.0 to Pypi, nor had we published 0.2.0 schemas to the schemastore, so this moves us back to 0.2.0 for publishing 0.2.0 to pypi and schemastore.

Related PRs:
https://github.com/EveryVoiceTTS/HiFiGAN_iSTFT_lightning/pull/51
https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/114
https://github.com/EveryVoiceTTS/DeepForcedAligner/pull/36


README.md
```diff
@@ -93,6 +93,44 @@ Have a look at [Contributing.md](Contributing.md) for the full details on the
...
+   - **Patch** version bump (bug fixes, small changes).
+
+2. **Update Version Files**
+   - Update the `everyvoice._version` file to reflect the new version.
+   - Keep all `submodule._version` files in sync with this version, **except** for the `wav2vec2` aligner submodule (which can be installed separately).
+   - Commit the resulting changes, including all submodules.
+
+3. **Update Schema (for Major/Minor bumps)**
+   If you bumped a **major** or **minor** version:
+   - Run `everyvoice update-schema`.
```
joanise, line 113: I would add `(You may need to delete some existing schema files if you get an error message.)` or something to that effect. Our process is not actually obvious except to you and I, because you just get an error message that effectively says "no can do" without giving guidance or fixing it. -- But maybe we want to keep the group of people doing this small?

README.md
```diff
@@ -93,6 +93,44 @@ Have a look at [Contributing.md](Contributing.md) for the full details on the
...
+
+4. **Open a Pull Request**
+   - Create a PR with your changes.
+   - Wait for tests to pass and for the PR to be merged into `main`.
+
+5. **Tag the Release**
+   After merging:
+   ```bash
+   git tag 'v0.X.X'
+   git push --tags
```
joanise, line 124: Please don't use `git push --tags`, that's dangerous. It's usually correct, but I would never do this, just in case I created tags locally, or I fetched a tag in the past that later got deleted.

There is no command that will tell you that a tag you have locally doesn't exist on origin, and this will blindly push them all.

Instead, I suggest always using `git push 'v0.X.X'` with a narrow command pushing just one tag, and more importantly having this documentation here use that form.

README.md
```diff
@@ -93,6 +93,44 @@ Have a look at [Contributing.md](Contributing.md) for the full details on the
...
+   - Commit the resulting changes.
+
+4. **Open a Pull Request**
+   - Create a PR with your changes.
+   - Wait for tests to pass and for the PR to be merged into `main`.
+
+5. **Tag the Release**
+   After merging:
+   ```bash
+   git tag 'v0.X.X'
```
joanise, line 123: Better: `git tag -a -m v0.X.X v0.X.X`

Annotated tags keep a trace of who created them and when, whereas lightweight tags are information-less pointers. Especially for creating release tags, it's important to have the -a and -m.

More notes on that at https://git-scm.com/book/en/v2/Git-Basics-Tagging if you care to read on it.

joanise: And a bit of nitpicking: 0.x.x implies rigidity that's not real, I would say X.Y.Z, but this change I don't actually care about.

joanise (commented):
Sorry, some more small changes, in the documentation. 

Otherwise I we're in business now.

README.md
roedoejet: Yes, you're right, but that doesn't mean we have to keep the documentation obscure. I just added a ruleset that prevents non-admin's from pushing v* tags.

README.md
roedoejet: noted!

README.md
roedoejet: cool, thanks! I changed to X Y Z

Issue #673, opened 2025-06-12
Implement text splitting for all synthesize and demo output formats


joanise:
### Description & Motivation

Once text splitting is implemented and supports merging the .wav outputs, the remaining output formats also need to be supported:
 - [x] spec
 - [ ] textgrid
 - [ ] readalong-xml
 - [ ] readalong-html

PR #697, opened 2025-07-03
Text Splitting Main Branch PR

### LINKED ISSUE #658: Integrate the splitter into EveryVoice synthesis

### LINKED ISSUE #691: Make Text Chunking Optional
### Description & Motivation

Text chunking should be set by default to `true` in TextConfig. User should be able to turn off chunking by using the -c flag when synthesizing from CLI or Demo.

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

jordanxlau:
### PR Goal? 

Allow chunking of texts when synthesizing audio with FastSpeech2, to better match longer audio inputs with training data.

More changes will be added to improve this feature, but this PR is for the **_minimum viable product_**. For example, text splitting is only available for output types wav and spec (a warning occurs when synthesizing another output type). Support for the other file outputs will come later.

### Fixes? 

This is a combination of 3 separate issues:
Fixes #691
Fixes #658
Solves 1/4 of #673 

### Feedback sought? 

Any

### Priority? 

high: I would like this to be merged sooner than later to prevent this large branch from becoming stale.

### Tests added? 

Yes

### How to test? 

- Perform all kinds of different text to wav synthesis (using demo, from filelist, from -t, using -c to turn splitting on and off).
- Also try to synthesize with --output-type ≠ wav and confirm that the warning message occurs.

You can see what splitting was performed by viewing `Processing text: []` (outputted to `stderr`).

### Confidence? 

All of these individual commits have been reviewed already, so I would put high, but this change has many moving parts so I am only **medium** confident.

### Version change? 

This is a **Minor** change, as it is fairly significant.

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/121

### DIFF
--- a/everyvoice/base_cli/helpers.py
+++ b/everyvoice/base_cli/helpers.py
@@ -19,7 +19,6 @@ from loguru import logger
 from pydantic import ValidationError
 from tqdm import tqdm
 
-from everyvoice.config.text_config import TextConfig
 from everyvoice.config.type_definitions import TargetTrainingTextRepresentationLevel
 from everyvoice.exceptions import InvalidConfiguration
 from everyvoice.model.e2e.config import EveryVoiceConfig
@@ -77,7 +76,6 @@ def load_config_base_command(
         type[EveryVoiceConfig],
         type[FastSpeech2Config],
         type[HiFiGANConfig],
-        type[TextConfig],
     ],
     # Must include the above in model-specific command
     config_args: List[str],
@@ -96,7 +94,6 @@ def load_config_base_command(
             EveryVoiceConfig,
             FastSpeech2Config,
             HiFiGANConfig,
-            TextConfig,
         ):
             try:
                 config = config_type.load_config_from_path(  # type: ignore[attr-defined]
--- a/everyvoice/base_cli/helpers.py
+++ b/everyvoice/base_cli/helpers.py
@@ -9,7 +9,6 @@ import json
 import os
 import tempfile
 import textwrap
-from enum import Enum
 from pathlib import Path
 from pprint import pformat
 from typing import List, Optional, Union
@@ -20,6 +19,7 @@ from loguru import logger
 from pydantic import ValidationError
 from tqdm import tqdm
 
+from everyvoice.config.text_config import TextConfig
 from everyvoice.config.type_definitions import TargetTrainingTextRepresentationLevel
 from everyvoice.exceptions import InvalidConfiguration
 from everyvoice.model.e2e.config import EveryVoiceConfig
@@ -37,6 +37,7 @@ from everyvoice.model.vocoder.HiFiGAN_iSTFT_lightning.hfgl.dataset import (
     HiFiGANDataModule,
 )
 from everyvoice.model.vocoder.HiFiGAN_iSTFT_lightning.hfgl.model import HiFiGAN
+from everyvoice.utils import update_config_from_cli_args
 
 MODEL_CONFIGS = [FastSpeech2Config, HiFiGANConfig]
 
@@ -76,12 +77,12 @@ def load_config_base_command(
         type[EveryVoiceConfig],
         type[FastSpeech2Config],
         type[HiFiGANConfig],
+        type[TextConfig],
     ],
     # Must include the above in model-specific command
     config_args: List[str],
     config_file: Path,
 ):
-    from everyvoice.utils import update_config_from_cli_args
 
     try:
         config = model_config.load_config_from_path(config_file)
@@ -95,6 +96,7 @@ def load_config_base_command(
             EveryVoiceConfig,
             FastSpeech2Config,
             HiFiGANConfig,
+            TextConfig,
         ):
             try:
                 config = config_type.load_config_from_path(  # type: ignore[attr-defined]
@@ -367,5 +369,9 @@ def train_base_command(
                 trainer.fit(model_obj, data, ckpt_path=tmp.name)
 
 
-def inference_base_command(_name: Enum):
-    pass
+def inference_base_command(
+    model: FastSpeech2,
+    config_args: List[str],
+):
+    config = model.config
+    update_config_from_cli_args(config_args, config)
--- a/everyvoice/base_cli/interfaces.py
+++ b/everyvoice/base_cli/interfaces.py
@@ -86,31 +86,6 @@ def train_base_command_interface(
 
 
 def inference_base_command_interface(
-    config_file: Path = typer.Argument(
-        ...,
-        exists=True,
-        dir_okay=False,
-        file_okay=True,
-        help="The path to your model configuration file.",
-    ),
     config_args: List[str] = typer.Option(None, "--config", "-c"),
-    accelerator: str = typer.Option(
-        "auto",
-        "--accelerator",
-        "-a",
-        help="Uses PyTorch Lightning Accelerators: https://pytorch-lightning.readthedocs.io/en/stable/extensions/accelerator.html",
-    ),
-    devices: str = typer.Option(
-        "auto", "--devices", "-d", help="The number of GPUs on each node"
-    ),
-    nodes: int = typer.Option(
-        1, "--nodes", "-n", help="The number of nodes on your machine"
-    ),
-    strategy: str = typer.Option(
-        "ddp",
-        "--strategy",
-        "-s",
-        help="The strategy for data parallelization: https://pytorch-lightning.readthedocs.io/en/stable/accelerators/gpu_intermediate.html",
-    ),
 ):
     pass
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -54,7 +54,7 @@ from everyvoice.model.vocoder.HiFiGAN_iSTFT_lightning.hfgl.cli import (
 )
 from everyvoice.model.vocoder.HiFiGAN_iSTFT_lightning.hfgl.cli import train as train_hfg
 from everyvoice.run_tests import SUITE_NAMES, run_tests
-from everyvoice.utils import load_config_from_json_or_yaml_path, spinner
+from everyvoice.utils import spinner
 from everyvoice.wizard import (
     PREPROCESSING_CONFIG_FILENAME_PREFIX,
     SPEC_TO_WAV_CONFIG_FILENAME_PREFIX,
@@ -786,8 +786,7 @@ def g2p(
     from everyvoice.text.phonemizer import get_g2p_engine
 
     if config:
-        kwargs = load_config_from_json_or_yaml_path(config)
-        text_config: TextConfig = TextConfig(**kwargs)
+        text_config: TextConfig = TextConfig.load_config_from_path(config)
         print(
             f"Config contains custon G2P Engines: {text_config.g2p_engines}",
             file=sys.stderr,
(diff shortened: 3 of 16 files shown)


everyvoice/config/text_config.py
```diff
@@ -174,6 +179,11 @@ class TextConfig(ConfigModel):
         description="User defined or external G2P engines.\nSee https://github.com/EveryVoiceTTS/everyvoice_g2p_template_plugin to implement your own G2P.",
         examples=["""{"fr": "everyvoice_plugin_g2p4example.g2p"}"""],
     )
+    split_text: bool = Field(
+        True,
+        title="Split Text",
+        description="Whether or not to perform text splitting (also referred to as text chunking) at inference time.",
```
roedoejet, line 185: Let's be even more descriptive here: "Whether or not to perform text splitting (also referred to as text chunking) at inference time. Instead of synthesizing an entire utterance, it will split the utterance into smaller chunks and re-combine them after synthesis. This can lead to more natural synthesis for long-form (i.e. paragraph) synthesis."

everyvoice/base_cli/helpers.py
```diff
@@ -76,12 +77,12 @@ def load_config_base_command(
         type[EveryVoiceConfig],
         type[FastSpeech2Config],
         type[HiFiGANConfig],
+        type[TextConfig],
```
roedoejet, line 80: we shouldn't actually be passing the text config here. better to pass the model config and access the text config from that (i.e. model.text) since we might want to edit other things besides the text config.

everyvoice/base_cli/helpers.py
```diff
@@ -20,6 +19,7 @@
 from pydantic import ValidationError
 from tqdm import tqdm
 
+from everyvoice.config.text_config import TextConfig
```
roedoejet, line 22: this can be removed (see explanation below)

everyvoice/base_cli/helpers.py
```diff
@@ -95,6 +96,7 @@ def load_config_base_command(
             EveryVoiceConfig,
             FastSpeech2Config,
             HiFiGANConfig,
+            TextConfig,
```
roedoejet, line 99: this can be removed, see explanation above.

everyvoice/base_cli/interfaces.py
```diff
@@ -95,32 +95,6 @@ def train_base_command_interface(
...
-    config_file: Path = typer.Argument(
-        ...,
-        exists=True,
-        dir_okay=False,
-        file_okay=True,
-        help="The path to your model configuration file.",
-        shell_complete=complete_path,
-    ),
     config_args: List[str] = typer.Option(None, "--config", "-c"),
-    accelerator: str = typer.Option(
```
roedoejet, line 107: Hm, can you actually make an issue to refactor these fields (accelerator, devices, nodes) back into the helper and interface here? I think these should maybe be shared as opposed to buried in the FS2 code as they currently are.

everyvoice/cli.py
```diff
@@ -587,6 +591,7 @@ def test(suite: TestSuites = typer.Argument("dev")):
 
 
 @app.command()
+@merge_args(inference_base_command_interface)
```
roedoejet, line 594: Hm, right, I guess it's fine to have the inference base command also add to the demo. nice.

everyvoice/config/text_config.py
```diff
@@ -227,3 +237,11 @@ def load_g2p_engines(self) -> Self:
             logger.info(f"Adding G2P engine from `{name}` for `{lang_id}`")
 
         return self
+
+    @staticmethod
+    def load_config_from_path(path: Path) -> "TextConfig":
```
roedoejet, line 242: good addition!

everyvoice/demo/app.py
```diff
@@ -91,12 +91,12 @@ def synthesize_audio(
         filelist=None,
         filelist_data=None,
         teacher_forcing_directory=None,
-        batch_size=1,
+        batch_size=16,
```
roedoejet, line 94: I guess we should actually get this from the model. `text_to_spec_model.config.training.batch_size` or at least configurable from the cli

everyvoice/base_cli/helpers.py
```diff
@@ -367,5 +369,9 @@ def train_base_command(
...
 
 
-def inference_base_command(_name: Enum):
-    pass
+def inference_base_command(
+    model: FastSpeech2,
+    config_args: List[str],
+):
+    config = model.config
+    update_config_from_cli_args(config_args, config)
```
roedoejet, line 377: Doesn't this have to return the config? Just updating the config here makes the inference base command a little light - I would prefer to see some other things relevant to inference, but as I mentioned in the other comment, that can wait for a new issue.

everyvoice/demo/app.py
```diff
@@ -207,6 +208,11 @@ def create_demo_app(
         device
     )
     model.eval()
+
+    from everyvoice.base_cli.helpers import inference_base_command
+
+    inference_base_command(model, **kwargs)
```
roedoejet, line 214: shouldn't this be returning the updated config? Or I guess it updates it in-place? I think what is happening here is a little opaque

everyvoice/run_tests.py
```diff
@@ -5,9 +5,9 @@
 
 import argparse
 import importlib
-import os
 import re
 import sys
+from os.path import dirname
```
roedoejet, line 10: this is fine, but not a necessary re-factor I don't think.

everyvoice/tests/test_cli.py
```diff
@@ -401,6 +402,22 @@ class WrongEnum(str, enum.Enum):
                 )
             self.assertIn("Unknown output format 'foo'", str(cm.exception))
 
+    def test_g2p(self):
```
roedoejet, line 405: did you add this? I'm not sure why this g2p code is in this PR

everyvoice/tests/test_text.py
```diff
@@ -561,3 +562,40 @@ def test_all_except_punctuation(self):
         self.assertSetEqual(
             symbols.all_except_punctuation, {"a", "b", "X", "Y", "Z", "<SIL>"}
         )
+
+
+class TestTextSplit(TestCase):
+    def test_strong_boundary(self):
```
roedoejet, line 568: nice, clear unit tests - excellent!

everyvoice/tests/test_text.py
```diff
@@ -561,3 +562,40 @@ def test_all_except_punctuation(self):
...
+        a = "There are approximately 70 Indigenous languages spoken in Canada!"
+        b = "Among these, there are 10 distinct language families."
+        c = "As a consequence of the residential school system and other policies of cultural suppression,"
+        d = "the majority of these languages now have fewer than 500 fluent speakers remaining."
+        e = "Most fluent speakers are elderly."
+        text = a + " " + b + " " + c + " " + d + " " + e
+        self.assertEqual([a + " " + b, c + " " + d, e], chunk_text(text, 75, 1000))
+
+    def test_normalization(self):
+        a = 'Welcome to the "EveryVoice" Documentation! Please read the background section below.'
```
roedoejet, line 590: beautiful tests!

everyvoice/text/textsplit.py
```diff
@@ -0,0 +1,99 @@
...
+    Split text into chunks of approximately `desired_length`,
+    trying to preserve sentence boundaries (!?.), with commas as lower-priority fallbacks.
+    """
+    # Validate arguments
+    assert desired_length < max_length
+
+    # Normalize input
+    text = re.sub(r"\n\n+", "\n", text)
+    text = re.sub(r"\s+", " ", text)
+    text = re.sub(r"[“”]", '"', text)
```
roedoejet, line 22: let's remove this quote normalization line

roedoejet: I think that's a vestige of my previous blizzard experiment

roedoejet (changes requested):
A few more requested changes here - but great work @jordanxlau ! it's looking great

everyvoice/base_cli/helpers.py
joanise: Yeah, we had discussions about it, and it's true it ended up quite light, but it's the thing we wanted to factor out from all the inference commands instead of adding it everywhere.

everyvoice/base_cli/interfaces.py
joanise: That's a good plan. The issue could include all these deleted interface declarations as documentation for whoever might implement this in the future.

everyvoice/demo/app.py
joanise: We struggled with this one when Jordan first saw that only the first chunk ever got looked at, because of that formerly hard-coded `batch_size=1`.

I'm not sure we want to use the training batch size, because we will typically use the big computing resources at training time, but potentially much smaller computing resources at inference time. Take it from the CLI would make sense, with a default that is typically OK. Now that Jordan implemented re-joining chunks across batches, that default could be lower. Aidan, do you have an intuition on what is reasonable here on typical inference computing resources?

everyvoice/run_tests.py
joanise: I didn't like writing `os.path.dirname(os.path.dirname(__file__))` so that was my choice.

joanise: Given how many instances there were, I got annoyed and changed the import.

everyvoice/tests/test_cli.py
jordanxlau: It's from the commit `test(refactor): test g2p cli command`. Basically, `load_config_from_path` was not implemented for `TextConfig`. I refactored it for consistency.

everyvoice/demo/app.py
joanise: Decision: rename this function `apply_shared_inference_parameters()`, and call the interface `apply_shared_inference_parameters_interface`

everyvoice/base_cli/interfaces.py
joanise: See #699

everyvoice/base_cli/interfaces.py
```diff
@@ -95,32 +95,6 @@ def train_base_command_interface(
...
 def inference_base_command_interface(
-    config_file: Path = typer.Argument(
-        ...,
-        exists=True,
-        dir_okay=False,
-        file_okay=True,
-        help="The path to your model configuration file.",
-        shell_complete=complete_path,
-    ),
     config_args: List[str] = typer.Option(None, "--config", "-c"),
```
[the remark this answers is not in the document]
joanise: See #700

everyvoice/demo/app.py
joanise: See #701

roedoejet (approved):
Great - thanks @jordanxlau !

PR #702, opened 2025-07-04
feat: cli rename speaker in checkpoint

### LINKED ISSUE #620: Model demo should be able to specify a different display name for speakers
### Description & Motivation

For example, we just found out that a speaker named `Doris` in the model should actually be named `Dorris` but in order to change this we currently need to manually edit the model checkpoint. We should instead be able to edit this from somewhere else.

### Pitch

I'm not totally sure where it makes sense to edit this. Command line? Some sort of configuration file?

### Alternatives

_No response_

### Additional context

_No response_

deltork:
### PR Goal? 
Adding the ability to change the speaker name in checkpoint

### Fixes? 
Fixes #620

### Feedback sought? 
basic check

### Priority? 
low

### Tests added? 
Test added

### How to test? 
from the cli 
`everyvoice inspect-checkpoint`

Get the speaker name

` everyvoice rename-speaker  MODEL_PATH OLD_SPEAKER_NAME  NEW_SPEAKER_NAME`

`everyvoice inspect-checkpoint` 

to see the new speaker name

### Confidence? 
100%

### DIFF
--- a/everyvoice/base_cli/checkpoint.py
+++ b/everyvoice/base_cli/checkpoint.py
@@ -176,3 +176,49 @@ def inspect(
                                 headers=["LayerName", "Number of Parameters"],
                             )
                         )
+
+
+@app.command()
+def rename_speaker(
+    model_path: Path = typer.Argument(
+        ...,
+        exists=True,
+        dir_okay=False,
+        file_okay=True,
+        help="The path to your model checkpoint file.",
+    ),
+    old_speaker_name: str = typer.Argument(
+        ...,
+        help="The name of the speaker to rename.",
+    ),
+    new_speaker_name: str = typer.Argument(
+        ...,
+        help="The new name for the speaker.",
+    ),
+):
+    """
+    Rename a speaker in the checkpoint's parameters.
+    """
+    ckpt = load_checkpoint(model_path, minimal=False)
+
+    if (
+        "hyper_parameters" in ckpt
+        and "speaker2id" in ckpt["hyper_parameters"]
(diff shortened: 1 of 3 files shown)


roedoejet (changes requested):
Thanks for this @deltork ! I think I had something different in mind though - instead of _renaming_ the speakers, I think I was imagining some way of passing a display name hash to the demo. So something like:

{"doris": "Dorris Montour"} where the key is the model's representation and the value is the display value. I think the challenge here is that there are at least three different reasons to access the name 1) the model needs a name defined in the CLI during inference or in the filelist provided for synthesis 2) the demo needs to display a name in the dropdown, and 3) the demo needs a slug to use for the API.

Maybe we can meet to discuss the possibilities here? I think we could actually keep this PR since it's a useful function. Maybe we could also update it to remove a speaker (i.e. remove their name from the list and zero-out the associated speaker embedding weights). In this case, I think it might be good to nest these commands under an `edit-checkpoint` command, like `everyvoice edit-checkpoint rename-speaker` and `everyvoice edit-checkpoint delete-speaker` or something like that. Anyways, let's chat at the meeting (or some other time, lmk!)

roedoejet:
@joanise suggests `everyvoice checkpoint ...` for group. So, `everyvoice checkpoint rename-speaker` or `everyvoice checkpoint inspect` etc...

roedoejet:
As discussed, this now relates to: https://github.com/EveryVoiceTTS/EveryVoice/issues/708

roedoejet (approved):
Looks good @deltork - we've got all the future upgrades described in #708 #714 and #715

PR #704, opened 2025-07-09
Several small changes not worth each their own PR

joanise:
### PR Goal? 

 - The way I use `silence_c_stderr` quite extensively to hide unit test logs has a problem when you make a change that causes a test to fail: you can't tell what went wrong! Now you can use `run_tests.py -v` to see those logs again. Or use `EVERYVOICE_VERBOSE_TESTS=1 python -m unittest ...` to enable these verbose logs when running a specific part of the test suites.
 - In CI, when tests fails, systematically run them again with this verbose switch, so we can see the problem in the CI logs too.
 - Update the documentation on how to run unit tests
 - The hidden command `everyvoice test` can only run if you installed from source - give an error message to that effect if you installed everyvoice from PyPI.
 - Add the -h synonym to --help for subsample.py.

### Fixes? 

This works stems from Jordan having problems with his PR failing CI without any kind of indication of what's wrong, because the helpful error logs that got generated were suppressed.

### Priority? 

low

### Tests added? 

it's only tests! :) 

### How to test? 

Compare the output of these two commands:
```
everyvoice/run_tests.py
everyvoice/run_tests.py --verbose
```

Compare the output of these two commands:
```
python -m unittest everyvoice.tests.test_cli
EVERYVOICE_VERBOSE_TESTS=1 python -m unittest everyvoice.tests.test_cli
```

Note especially here, when tests pass, it's nice to just have a line of dots on the screen (first instance) but on failure the logs shown in the second instance might help figure out what was wrong.

I suppose an even better solution would be to capture the logs and print them when a test case fails, but I'm not up to writing that automated logic at the moment...

### Confidence? 

high

### Version change? 

no

### Related PRs? 

not yet, but maybe I'll do the same CI changes in the submodules


README.md
```diff
@@ -148,4 +148,9 @@ Many thanks to:
 
 ## Tests
 
-Run unit tests by `python -m unittest tests/test_configs.py` or suites of tests by running `everyvoice test dev` if you have the package installed interactively.
+There are many ways to run the unit tests, if you installed EveryVoice from source:
+ - Run all the dev tests: `everyvoice/run_tests.py dev` or `everyvoice test dev`
```
roedoejet, line 152: don't you mean `python everyvoice/run_tests.py dev` ?

README.md
```diff
@@ -148,4 +148,9 @@ Many thanks to:
...
+ - Run the tests with verbose logs: `everyvoice/run_tests.py --verbose dev`
+ - Show the names of the other suites you can run: `everyvoice/run_tests.py -h`
+ - Run all the tests in one test file: `python -m unittest everyvoice/tests/test_<somefilename>.py`
+ - Run one specific test case: `python -m unittest everyvoice.tests.<filename>.<class_name>.<function_name>`
```
roedoejet, line 156: good idea to add this in, even though it's standard

everyvoice/cli.py
```diff
@@ -566,9 +566,17 @@ def new_project(
...
-def test(suite: TestSuites = typer.Argument("dev")):
+def test(suite: TestSuites = typer.Argument("dev")):  # pragma: no cover
     """Run a test suite"""
-    run_tests(suite.value)
+    try:
+        import everyvoice.tests  # noqa: F401
+
+        run_tests(suite.value)
+    except ModuleNotFoundError:
+        print(
```
roedoejet, line 576: nice

roedoejet (approved):
nice improvements here, thanks @joanise !

README.md
joanise: Nope, `run_tests.py` has the shebang line needed to make it directly executable. It works on my machine, and it should work cleanly on any *nix-based OS. The only case where it's not reliable is on the Windows command line. So yes, I mean `everyvoice/run_tests.py dev`.

PR #709, opened 2025-07-16
[WIP] feat: add detailed missing audio file reporting to preprocessor

MENGZHEGENG:
- Add missing_files_list to track missing audio file paths during preprocessing
- Enhance report() method to include detailed missing files section
- Save missing files to dedicated missing_files.txt file in output directory
- Track missing files at all detection points in preprocessing pipeline
- Maintain backward compatibility while improving debugging experience


joanise (commented):
for unit testing this PR, the example to start from and adapt are more like `test_gotta_do_audio_first` and `test_empty_preprocess`. There's also `test_incremental_preprocess` that's a bit more complex but also related.

They all call `Preprocessor.preprocess()`, and they all use `get_simple_config()` to do a lot of the work needed to be able to call that function. If you can reuse that, it will be a lot less work.

You would have to adapt those example to change the filelist to have missing audio files. You can look at `test_empty_preprocess` too see how it changes the filelist to create another error situation.

And then you should be able to have an assertion that the missing files text file exists, and maybe even open and read it and make sure it has the contents you expect.

A negative test would be useful too: call `preprocess()` with the default filelist from `get_simple_config()` and assert that the missing files text file does not exist on the file system.

everyvoice/preprocessor/preprocessor.py
```diff
@@ -352,7 +353,18 @@ def report(self, tablefmt="simple", audio_only=False):
             table = audio_1 + audio_2
         else:
             table = audio_1 + others + audio_2
-        return tabulate(table, headers, tablefmt=tablefmt)
+
+        report_text = tabulate(table, headers, tablefmt=tablefmt)
+
+        # Add missing files section if any exist
+        if self.missing_files_list:
```
roedoejet, line 360: Nice and elegant solution.

everyvoice/preprocessor/preprocessor.py
```diff
@@ -1042,6 +1056,19 @@ def preprocess(  # noqa: C901
                     with open(self.save_dir / "summary.txt", "w", encoding="utf8") as f:
                         f.write(report)
                         f.write("\n")
+
+                    # Save missing files list to a separate file
+                    if self.missing_files_list:
+                        with open(
+                            self.save_dir / "missing_files.txt", "w", encoding="utf8"
```
roedoejet, line 1063: I think maybe this should be time stamped or somehow named according to the run, since this will just overwrite the file otherwise.

roedoejet (changes requested):
Just a small suggestion to add a timestamp or some sort of run identifier to the `missing_files.txt` filename so that it doesn't overwrite the file on subsequent runs.

everyvoice/tests/test_preprocessing.py
```diff
@@ -688,6 +688,191 @@ def test_no_speaker(self):
...
+            tmpdir = Path(tmpdir)
+            preprocessor.save_dir = tmpdir
+
+            # Simulate the preprocess method's report generation
+            if "audio" in ("audio",):
+                report = "Here is a report:\n" + preprocessor.report()
+                if not preprocessor.counters.value("duration"):
+                    report += "\n\nWARNING: No audio files were processed."
+
+            self.assertIn("WARNING: No audio files were processed.", report)
```
joanise, line 874: Other tests in this suite are great, but I don't understand this one: it's asserting a message that you add here in the unit test, rather than one generated by the preprocessor. Maybe I am misunderstanding something?

everyvoice/preprocessor/preprocessor.py
```diff
@@ -1121,6 +1133,21 @@ def preprocess(  # noqa: C901
                             f.write("=" * 50 + "\n")
                             for multichannel_file in self.multichannel_files_list:
                                 f.write(f"{multichannel_file}\n")
+
+                    # Save missing files list to a separate file with timestamp
+                    if self.missing_files_list:
```
roedoejet: nice! thanks @MENGZHEGENG

PR #710, opened 2025-07-18
Wizard refactorings in preparation for the custom g2p enhancements

joanise:
### PR Goal? 

The PR for custom g2p questions involved a lot of stuff, and is going to be a mess to review.

The help alleviate the situation, I have created this PR which is only refactorings of the wizard, that I did for the purpose of the custom g2p improvements, but which can stand alone by themselves and therefore can be reviewed separately.

### Feedback sought? 

general review and sanity checking.

At times, you will wonder "why bother?" for some changes. By default, assume "because it supported writing the custom g2p questions."

### Priority? 

normal

### Tests added? 

This code was already well covered by tests. The next PR will have lots of new tests, but not this one.

### How to test? 

I think if CI passes, it's all good, to be honest. But you can also run the Wizard and see that there are no surprises.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

This PR prepares the way for #711

PR #711, opened 2025-07-18
Dev.ej/custom g2p wiz

### LINKED ISSUE #645: Augment the wizard to handle user-defined G2P engines
### Description & Motivation

It was decided to limit the scope of #635 & #644 to create a new issue to add the necessary code in the wizard to handle the case where the user wants to use its own custom G2P engine.

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

joanise:
### PR Goal? 

This is my second big PR related to custom g2p in the wizard: this one adds the questions and handles it all, plus testing.

Known bug: if you have a given lang that is used in more than one dataset in the same project, and you set the custom g2p on any other than the first where that language occurs, the earlier datasets won't have been processed with the custom g2p.
I am deferring fixing this issue to its own PR as it entails another significant round of refactoring that I want to do later and have reviewed separately.
Fixing that bug should require no user-visible changes, so the wizard enhancements can be tested and reviewed before that gets fixed.

### Out of scope

 - fixing the bug described above -- ticket created: #712
 - documentation, that's for #354 
 - I might want to handle an old resume files automatically, since the adjustments to run through the current wizard from an old resume file are deterministic. But if I do that, it will be in a separate PR.

### Fixes? 

Fixes #645

### Feedback sought? 

Here I am hoping for good testing, making sure the wizard works well in various scenarios and doesn't break. @marctessier ?

Is the series of questions intuitive or at least easy enough to use?

And @roedoejet I'm hoping for a good code review, it's quite involved.

### Priority? 

normal

### Tests added? 

yup, lots, with high coverage.

### How to test? 

Run the wizard and play with the custom g2p questions, for datasets with and without a language column, with a custom code or not, and using your own g2p or not.

You can use https://github.com/EveryVoiceTTS/everyvoice_g2p_template_plugin to get ideas on creating more custom g2p functions, but in testing I have been using `everyvoice.tests.g2p_engines.valid` every time I wanted to set a custom one.

Entering bad functions on that custom g2p engine question is also a good thing to test, all problems you can imagine should give a friendly error: module does not exist, function does not exist, wrong signature, and who knows what else.

### Confidence? 

decent

### Version change? 

Yes. We need to at least bump the patch since the wizard becomes incompatible with previous instances if resume files.

But this does not affect the schemas, so bumping the patch is enough.

### Related PRs? 

Builds on top of #710

### DIFF
--- a/everyvoice/tests/test_wizard.py
+++ b/everyvoice/tests/test_wizard.py
@@ -32,6 +32,7 @@ from everyvoice.tests.stubs import (
     patch_menu_prompt,
     patch_questionary,
 )
+from everyvoice.text.phonemizer import AVAILABLE_G2P_ENGINES
 from everyvoice.wizard import StepNames as SN
 from everyvoice.wizard import basic, dataset, prompts, utils
 from everyvoice.wizard.main_tour import get_main_wizard_tour
@@ -623,6 +624,23 @@ class WizardTest(TestCase):
         with patch_menu_prompt(0):  # 0 is "characters"
             text_representation_step.run()
         self.assertEqual(step.state["filelist_headers"][2], "characters")
+
+        text_processing_step = find_step(SN.text_processing_step, tour.steps)
+        # 0 is lowercase, 1 is NFC Normalization, select both
+        with monkeypatch(dataset, "tqdm", lambda seq, desc: seq):
+            with patch_menu_prompt([0, 1]):
+                text_processing_step.run()
+        # print(text_processing_step.state)
+        self.assertEqual(
+            text_processing_step.state["filelist_data_list"][2][2],
+            "cased nfd: éàê nfc: éàê",  # the "nfd: éàê" bit here is now NFC
+        )
+
+        self.assertEqual(
+            text_processing_step.state["filelist_data_list"][3][2],
+            "let us see if it collapses whitespace",
+        )
+
         speaker_step = find_step(SN.data_has_speaker_value_step, tour.steps)
         children_before = len(speaker_step.children)
         with patch_menu_prompt(0):  # 0 is "no"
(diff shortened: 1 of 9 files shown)


everyvoice/wizard/dataset.py
```diff
@@ -689,20 +727,198 @@ def effect(self):
...
+        )
+
+        _, g2p_langs_full = get_arpabet_langs()
+        self.language_names = [
+            g2p_langs_full.get(language, language) for language in self.language_codes
+        ]
+
+        self.current_engines = []
+        self.set_a_custom = []
+        for language in self.language_codes:
```
roedoejet, line 768: What if someone labelled two different datasets as `crk` and then realized that actually they should have labelled them `crk-east` and `crk-west` since they use different g2p engines - is there a way to edit that at this point? Or would the recommendation be to edit the config file?

everyvoice/wizard/dataset.py
```diff
@@ -689,20 +727,198 @@ def effect(self):
...
+            current = AVAILABLE_G2P_ENGINES.get(language, None)
+            if isinstance(current, CachingG2PEngine) or current == DEFAULT_G2P:
```
roedoejet, line 770: I really like this UX.

everyvoice/wizard/dataset.py
```diff
@@ -689,20 +727,198 @@ def effect(self):
...
+        self.language_name = language_name
+        self.current_engine = current_engine
+
+    def prompt(self):
+        if self.language_code == self.language_name:
+            lang_desc = self.language_name
+        else:
+            lang_desc = f"{self.language_name} ({self.language_code})"
+        rich_print(
+            Panel(
```
roedoejet, line 861: nice

everyvoice/wizard/dataset.py
```diff
@@ -888,6 +1101,21 @@ def validate(self, response):
...
+        self.saved_state = {
+            "model_target_training_text_representation": None,
+        }
+
+        # apply automatic conversions before extracting symbols
+        # - text processing
+        # - g2p conversion
+        # and get target training representation
+        self.state["model_target_training_text_representation"] = (
+            apply_automatic_text_conversions(
```
roedoejet, line 1113: But if we haven't defined the custom g2p yet, this wont run, right?

everyvoice/tests/test_wizard.py
```diff
@@ -623,6 +625,23 @@ def test_dataset_subtour(self):
...
+
+        text_processing_step = find_step(SN.text_processing_step, tour.steps)
+        # 0 is lowercase, 1 is NFC Normalization, select both
+        with monkeypatch(dataset, "tqdm", lambda seq, desc: seq):
+            with patch_menu_prompt([0, 1]):
+                text_processing_step.run()
+        # print(text_processing_step.state)
+        self.assertEqual(
+            text_processing_step.state["filelist_data_list"][2][2],
+            "cased nfd: éàê nfc: éàê",  # the "nfd: éàê" bit here is now NFC
```
roedoejet, line 637: I prefer using `e\u0301` syntax for this reason, but this is fine too

everyvoice/tests/test_wizard.py
```diff
@@ -652,14 +671,26 @@ def test_dataset_subtour(self):
 
         select_lang_step = language_step.children[0]
         with capture_stdout(), capture_stderr():
-            with patch_menu_prompt(15):  # some arbitrary language from the list
+            # We rely on Finnish g2p in assertions below, so find it in the list.
+            # 2 + index(fin) because in select_lang_step, [0] is "und", [1] is "custom",
```
roedoejet, line 675: thanks for this note - the 2 would have been cryptic otherwise.

roedoejet (approved):
I've taken a good look at this and it's very comprehensive. A couple minor questions and points of clarifications, but the testing is very extensive here and I think it looks ready to merge.

everyvoice/wizard/tour.py
```diff
@@ -321,7 +321,7 @@ def resume(self, resume_from: Path) -> Optional[Step]:
         # for new questions if possible, or else giving a warning explaining what needs
         # to be changed if auto upgrade is not possible.
         # Regression testing should warn us when such auto-upgrade code is required here.
-        compatible_since = Version("0.2.0a0")
+        compatible_since = Version("0.4.0.dev0")
```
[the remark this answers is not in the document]
joanise: No, because we have not published 0.4.0 yet.

PR #713, opened 2025-07-21
Group all custom g2p tests in one new test file

joanise:
### PR Goal? 

There are a growing number of tests related to custom g2p, that touch various parts of the code base but are all related.

Also, test_wizard is getting unwieldingly large.

So I propose to group all the custom g2p tests in one place.

This PR does that, and only that, it changes no tests, it only moves them around.

I want to isolate this to make review manageable, because combining this with any other PR will make that have a huge diff where it won't be clear what's new and what's just moved.

### Feedback sought? 

Sanity checking

### How to test? 

Just exercise unit tests before and after, and see the same number of tests run and passed.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

Builds on #711

### DIFF
--- a/everyvoice/run_tests.py
+++ b/everyvoice/run_tests.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 
-""" Organize tests into Test Suites
-"""
+"""Organize tests into Test Suites"""
 
 import argparse
 import importlib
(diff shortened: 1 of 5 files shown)


roedoejet (approved):
thanks for splitting this up! I agree it would have been harder to review otherwise. looks good!

PR #721, opened 2025-08-05
dev.jl/configure boundaries

### LINKED ISSUE #659: Integrate the text splitting with TextConfig

jordanxlau:
### PR Goal? 

Allow the user to configure text chunking boundaries in TextConfig (eg. adding an interrobang or ellipsis to the list of strong boundaries, or removing a colon from the list of weak boundaries).

### Fixes? 

Fixes #659 

### Feedback sought? 

Have I properly credited the text in `test_text.py`?
Are there enough unit tests (especially in the submodule)?

### Priority? 

Medium

### Tests added? 

Yes, see `test_text.py` and `test_cli.py`.

### How to test? 

Run commands like below, changing the default strong and weak boundaries. Preferably do this in a language like SENĆOŦEN or Cree, where punctuation symbols are different/used differently.

`everyvoice synthesize from-text ... -c text.strong_boundaries=!.`

You could also edit the text config .json or .yaml file, and train a new model with custom boundaries.

Make sure to test (perhaps with a breakpoint in `synthesize.py`) that the model splits text according to your new boundaries.

### Confidence? 

Medium

### Version change? 

patch

### Related PRs? 

See submodule PR: https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/125


everyvoice/config/text_config.py
```diff
@@ -139,9 +139,21 @@ def member_must_be_list_of_strings(self) -> "Symbols":
     ),
 ]
 G2P_Engines = Annotated[
-    dict[G2P_lang, G2P_py_module],
+    dict[Language, G2P_py_module],
     Field(description="Mapping from language id to g2p module"),
 ]
+LanguageBoundaries = Annotated[
+    dict[str, str],
```
jordanxlau, line 146: I intially tried to make an Enum called `BoundaryType` (that would be only either "strong" or "weak") but this caused unpickling errors on `torch.load()`. If you think an Enum would be super beneficial, I can attempt to deal with those errors.

PR #731, opened 2025-08-15
dev.jl/get lengths from data

### LINKED ISSUE #660: Get desired length and max-length from dataset
### Description & Motivation

The desired length is relative to the language and dataset. We should try and set this by inspecting the average utterance length in the dataset and suggest setting the desired length to that. We could maybe do this in preprocessing.

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

jordanxlau:
### PR Goal? 

Add calculation of character length and phone length to stats.json. These will be used in the submodule.

### Fixes? 

Fixes #660 

### Tests added? 

Yes: see [this commit](https://github.com/EveryVoiceTTS/EveryVoice/pull/731/commits/ce9c83d1fda578cc52b9a77d16b7cee048f36642).

### Version change? 

patch

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/126


joanise (approved):
Looks good.

PR #745, opened 2025-09-16
Dev.ej/refactor and cleanup

joanise:
Minor code cleanup:
 - multiple `self.assertIn()` calls were on multiple lines for no good reason (black didn't reformat back because of the trailing comma)
 - `Scaler` already implements `.append()`, we should not call `.data.append()`, but rather just `.append()`.


roedoejet (approved):
nice and simple - lgtm!

PR #747, opened 2025-09-19
Dev.ej/repair licensecheck

joanise:
### PR Goal? 

1) The quite some time now, the `licensecheck` call in the workflow has not been working, and thus our automated due diligence has been disabled. This PR primarily fixes that.

~2) While working on the PR, it's annoying how pre-commit and licensecheck, with no dependencies on the EV test suite, delay it and/or are delayed by it. So I refactored the workflow into three independent jobs which can start at the same time and provide their independent results faster.~ UPDATE: combining these changes makes review too hard, splitting this into a separate PR. #750

### Fixes? 

broken license check

### Priority? 

medium

### Tests added? 

it's all just tests

### How to test? 

look at the CI runs for this PR

### Confidence? 

high:
 - I have push commits that deliberately make license checking fail and see them fail at the right place
 - I will make sure a pre-commit failure has the intended effect shortly (once this PR is created)

### Version change? 

no

### Related PRs? 

none

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,18 +96,18 @@ jobs:
         with:
           preformatted: true
           message-path: import-message.txt
-      - name: Check licenses
-        run: |
-          pip install licensecheck --no-warn-conflicts
-          # Make sure we don't have or introduce dependencies with incompatible licenses.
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
lgtm - I like the improved error message and documentation

PR #750, opened 2025-09-22
perf(ci): split the test workflow into three parallel workflows

joanise:
### PR Goal? 

While working on #747, it was annoying how pre-commit and licensecheck, with no dependencies on the EV test suite, delay it and/or are delayed by it. So I refactored the workflow into three independent jobs which can start at the same time and provide their independent results faster.

Now the three independent parts of the workflow can run in parallel.

Also optimize licensecheck and pre-commit to use the default python as is, speeding up their launch.

And bring the pre-commit-ci action to the current latest version

EDIT: and use `awalsh128/cache-apt-pkgs-action@v1.5.3` to apt install packages with caching, giving us another significant action speed up.

### Fixes? 

how quickly CI results can become available

### Feedback sought? 

regular review

### Priority? 

low

### How to test? 

watch CI

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/.github/workflows/matrix-tests.yml
+++ b/.github/workflows/matrix-tests.yml
@@ -20,12 +20,13 @@ jobs:
       - uses: actions/checkout@v5
         with:
           submodules: recursive
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
-      - uses: FedericoCarboni/setup-ffmpeg@v2
-      - run: sudo apt-get install sox libsox-dev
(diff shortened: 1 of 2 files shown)


SamuelLarkin (approved):
LGTM
Should be also pin gh/actions to a sha1?  Not suggesting to do this in this PR .

joanise:
> Should be also pin gh/actions to a sha1? Not suggesting to do this in this PR .

I deliberately left those out, on the theory that GitHub exercises stronger control on the publication of their own actions, and in that case I'm happier getting the patches than pinning. But it remains a question open for discussion.

PR #758, opened 2025-10-21
Add support for dataset-specific data cleaners and replacement rules

joanise:
### PR Goal? 

Sometimes, different datasets will require different cleaners, e.g., a multi lingual system with both moh and str might only want to lowercase the moh dataset but not the str one. Add support for such a configuration.

### Fixes? 

Fixes #752
Fixes #592 
Fixes #359 

### Feedback sought? 

This touches a lot of stuff, so a good code review please and testing all the way to training.

### Priority? 

normal

### Tests added? 

yes

### How to test? 

Create a project with datasets requiring different cleaners, maybe also different to_replace rules. 
 - The wizard should create a configuration with global and dataset-specific cleaners -- the global ones will be the intersection of the dataset specific ones.
 - preprocessing will apply only the appropriate dataset specific cleaners for each dataset
 - synthesis will apply the language-specific cleaners, which are inferred from the dataset ones by the wizard

### Confidence? 

good

### Version change? 

Yes, will require a minor bump because I'm changing the schemas.

### Related PRs? 

no

PR #759, opened 2025-10-22
Code improvements while working on #758

joanise:
### PR Goal? 

To make #758 easier to review, I am splicing refactorings and other code improvements that are not directly related to that problem into a separate PR.

Generally, the idea is to refactor to reduce technical debt. Specifically:
 - Fix some type annotations that Pylance flagged: even though mypy didn't say anything, I fixed the ones where I think complying with Pylance makes better code
 - Silence some test suites - when you run the tests, there should really be no output at all except saying which tests ran and the result
 - Most annoying: when you use `Field(value)` with the default value specified by position, VSCode autocompletion and pylance don't process it correctly, you need to use `Field(default=value)` by keyword. I'm not that happy with making the change, but it's a developer quality of life improvement and I decided I wanted to do it. Neither MS nor Pydantic will budge on this and adapt to the other (as evidenced by some issues on relevant repos) and the answer is "just use `default=` by keyword if you want the two to play nice with each other", even though that should not be required.
 - Fix lots more errors flagged by `mypy --check-untyped-defs everyvoice`

### Priority? 

I want to merge this before #758

### How to test? 

just run unit tests

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/130
https://github.com/EveryVoiceTTS/HiFiGAN_iSTFT_lightning/pull/57

### DIFF
--- a/everyvoice/base_cli/callback.py
+++ b/everyvoice/base_cli/callback.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict
+from typing import Any
 
 import pytorch_lightning as pl
 from pytorch_lightning.callbacks import Callback
@@ -16,7 +16,7 @@ class ResetValidationDataloaderCallback(Callback):
         self,
         trainer: "pl.Trainer",
         pl_module: "pl.LightningModule",
-        checkpoint: Dict[str, Any],
+        checkpoint: dict[str, Any],
     ) -> None:
         batch_progress = trainer.fit_loop.epoch_loop.val_loop.batch_progress
         batch_progress.reset()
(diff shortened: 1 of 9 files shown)


everyvoice/preprocessor/preprocessor.py
```diff
@@ -445,22 +447,22 @@ def compute_stats(
...
-                        [len(line["characters"])], dtype=float
+                        [len(line["characters"])], dtype=torch.float64
                     )
                     char_length_scaler.append(char_length_data)
             if phone_length:
                 phone_length_scaler = Scaler()
                 for line in tqdm(filelist, desc="Gathering phone length values"):
-                    phone_length_data = torch.tensor([len(line["phones"])], dtype=float)
+                    phone_length_data = torch.tensor(
+                        [len(line["phones"])], dtype=torch.float64
```
roedoejet: same here

PR #767, opened 2025-11-24
Fix the ffmpeg-caused errors in EV CI

joanise:
### PR Goal? 

Fix CI so that ffmpeg-based test in the w2v submodule work.

### Fixes? 

For some reason, apt-get installed ffmpeg depends on libblas.so.3 which didn't get installed. It might have been possible to figure out how to install it correctly, but using Federico's ffmpeg action is very reliable, runs faster, and solves the problem.

For the record, the actual error message was:

> ffmpeg: error while loading shared libraries: libblas.so.3: cannot open shared object file: No such file or directory

### Feedback sought? 

mostly rubber stamping

### Priority? 

well, pretty high to allow for synching the submodule mains to the EV main

### Tests added? 

just fixing existing tests

### How to test? 

See that CI passes, both tests.yml and matrix tests

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/wav2vec2aligner/pull/26


roedoejet (approved):
looks great, I agree, better to use the published action - thanks @joanise !

PR #784, opened 2026-04-10
Dev.ej/deprecated typing

### LINKED ISSUE #718: Replace deprecated typing alias by their official collections.abc or builtin classes
### Bug description

The following typing aliases have been deprecated since Python 3.9.

 - typing.List
 - typing.Dict
 - typing.Callable
 - typing.Sequence
 - and many more, documented in https://docs.python.org/3/library/typing.html#deprecated-aliases

We should replace them by their current recommended equivalents:

 - list
 - dict
 - collections.abc.Callable
 - collections.abc.Sequence
 - etc.

While it's still safe to use either variant for the moment, [PEP 585](https://peps.python.org/pep-0585/) says:
> The deprecated functionality may eventually be removed from the typing module. Removal will occur no sooner than Python 3.9’s end of life, scheduled for October 2025.
October 2025 is coming soon so it would be good to adapt sooner than later if we want to be compatible with future versions of Python.

For wav2vec2aligner, where we declare support for Python 3.8, we could use `from __future__ import annotations` to enable subscripting `list[...]` and `dict[...]`.

### How to reproduce the bug

```python
# Sample code to reproduce the problem
```

### Error messages and logs

```
# Error messages and logs here please
```

### Environment

<details>
<summary>Current environment</summary>
<pre>
# Please paste the output of `everyvoice --diagnostic` here
# EveryVoice Diagnostic information
</pre>
</details>

### More info

_No response_

joanise:
### PR Goal? 

Remove deprecated typing aliases as described in #718

### Fixes? 

Fixes #718

### Tests added? 

Already tested

### How to test? 

Just run `pytest`

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/135
https://github.com/EveryVoiceTTS/HiFiGAN_iSTFT_lightning/pull/60

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -3,7 +3,7 @@
 contrib=contrib-title-conventional-commits
 
 # We don't require a body, just a title, even though a body is also a good idea
-ignore=body-is-missing
+ignore=body-is-missing,body-min-length
 
 [contrib-title-conventional-commits]
 # Specify allowed commit types. For details see: https://www.conventionalcommits.org/
--- a/everyvoice/base_cli/checkpoint.py
+++ b/everyvoice/base_cli/checkpoint.py
@@ -7,10 +7,9 @@ import warnings
 from collections import defaultdict
 from json import JSONEncoder
 from pathlib import Path
-from typing import Any
+from typing import Annotated, Any
 
 import typer
-from typing_extensions import Annotated
 
 from .interfaces import typer_file_argument
 
--- a/everyvoice/config/shared_types.py
+++ b/everyvoice/config/shared_types.py
@@ -3,12 +3,12 @@ Types shared among all EV modules, though not always super fast to load.
 Tiny type defs that load in milliseconds can go into type_definitions.py.
 """
 
-from collections.abc import Mapping, Sequence
+from collections.abc import Callable, Iterator, Mapping, Sequence
 from contextlib import contextmanager
 from contextvars import ContextVar
 from functools import cached_property
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterator, Tuple, Union
+from typing import Any, Union
 
 from pydantic import (
     BaseModel,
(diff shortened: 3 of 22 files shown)


SamuelLarkin (approved):
I've fixed a few more and pushed a commit.

When I run `ruff check --select UP006,UP035`, there is this one that I think should also be fixed but I would prefer that @joanise has a look at this first.  I wasn't sure if that was kept to be backward compatible or something.
```
UP006 [*] Use `list` instead of `typing.List` for type annotation
   --> everyvoice/config/text_config.py:180:9
    |
178 |     assert sig.return_annotation in (
179 |         list[str],
180 |         typing.List[str],
    |         ^^^^^^^^^^^
181 |     ), "G2P Engine's signature should return a list of strings"
    |
help: Replace with `list`

Found 1 error.
[*] 1 fixable with the `--fix` option.
```

joanise:
> I've fixed a few more and pushed a commit.
> 
> When I run `ruff check --select UP006,UP035`, there is this one that I think should also be fixed but I would prefer that @joanise has a look at this first. I wasn't sure if that was kept to be backward compatible or something.
> 
> ```
> UP006 [*] Use `list` instead of `typing.List` for type annotation
>    --> everyvoice/config/text_config.py:180:9
>     |
> 178 |     assert sig.return_annotation in (
> 179 |         list[str],
> 180 |         typing.List[str],
>     |         ^^^^^^^^^^^
> 181 |     ), "G2P Engine's signature should return a list of strings"
>     |
> help: Replace with `list`
> 
> Found 1 error.
> [*] 1 fixable with the `--fix` option.
> ```

Yeah, this one is required to stay, actually, because it's a user limitation otherwise: we want to allow people to write their custom g2p functions with either syntax. Previously, we only allowed `typing.List[str]`, so if I were to now only allow `list[str]` I would actually break any custom g2p function anyone has ever written, probably including our own g2p modules since we're still keeping it compatible with Python 3.8...

joanise:
`# noqa: UP006` allows me to keep that `typing.List` while not tripping the ruff checker, I'll use that.

PR #786, opened 2026-04-15
Better error messages for incompatible models

joanise:
### PR Goal? 

Two related changes to improve how we handle old and/or incompatible models.
 - `everyvoice checkpoint inspect` now uses the embedded `model_info` to directly target the right class, if possible, with the same fallback cascade as before if it doesn't work
 - starting the demo/app via `everyvoice demo` or a specific `app.py` script like we do on HF should produce more helpful error messages in cases of failures.

### Fixes? 

My hope is to reduce the troubleshooting efforts required next time we try to start a model with the wrong version of EV.

### Feedback sought? 

@SamuelLarkin the changes in `checkpoint.py` refactor your code, please review the code carefully

@deltork opinion on the error messages since you were troubleshooting the system that brought me to want to make these improvements.

### Priority? 

normal

### Tests added? 

yeah, not really. We don't have models saved in the repo accessible for CI, so none of these exception are exercised. I want to create a bank of old models created with various versions of EV for regression testing, but that's not a job for this PR.

### How to test? 

`everyvoice inspect` a variety of models and see slightly improved output, and especially no errors due to trying other kinds of models in the cases where the model_info field tells us what the right kind is.

### Confidence? 

high

### Version change? 

nah

### Related PRs? 

nope


everyvoice/config/text_config.py
```diff
@@ -102,7 +102,7 @@ def cannot_have_punctuation_in_symbol_set(self) -> "Symbols":
         for punctuation in self.punctuation.all:
             if punctuation in self.all_except_punctuation:
                 raise ValueError(
-                    f"Sorry, the symbol '{punctuation}' occurs in both your declared punctuation and in your other symbol set. Please inspect your text configuration and either remove the symbol from the punctuation or other symbol set."
+                    f"Sorry, the symbol '{punctuation}' occurs in both your declared punctuation and in your other symbol set. If you are training a model, please inspect your text configuration and either remove the symbol from the punctuation or other symbol set. If you are trying to use an existing model, it was probably trained with an older version of EveryVoice, please downgrade EveryVoice."
```
roedoejet, line 105: Is there no way to suggest a version?

roedoejet (approved):
A nice solid improvement on checkpoint inspection including some sensible refactoring. Lgtm, thanks @joanise !

everyvoice/config/text_config.py
joanise: Yeah, there probably is, but that might require a fair bit more work. I'll push it to a possible later improvement, but it would be a good idea.

PR #788, opened 2026-04-21
fix dependency problems with EV 0.3.0

joanise:


roedoejet (approved):
great. thanks @joanise

PR #790, opened 2026-04-24
Correctly find the IPA language code for sal-apa

### LINKED ISSUE #789: using `sal-apa` as language triggers `InvalidLanguageCode` on `sal-apa-ipa`
### Bug description

When you specify data in `sal-apa` and don't provide the `phones` column, the wizard will use `g2p` to convert to phones. In `everyvoice.text.phonemizer.CachingG2PEngine.__init__()`, we assume the IPA language code is `f"{lang_id}-ipa"`, but that's wrong for `sal-apa` and `oji.syl`.

joanise:
### PR Goal? 

Correctly find the IPA language code for `sal-apa`

### Fixes? 

Fixes #789

### Feedback sought? 

regular review

### Priority? 

high

### Tests added? 

yes

### How to test? 

run through the wizard with data in `sal-apa`, and see it pass by the language selection step (exception dumps there before this PR).

### Confidence? 

high

### Version change? 

no, but we're due

### Related PRs? 

https://github.com/NRC-ILT/g2p/pull/489

### DIFF
--- a/everyvoice/tests/test_custom_g2p.py
+++ b/everyvoice/tests/test_custom_g2p.py
@@ -95,6 +95,15 @@ class TestG2p(TestCase):
         with self.assertRaises(NotImplementedError):
             get_g2p_engine("boop")
 
+    def test_unusual_ipa_code(self):
+        # sal-apa goes to sal-ipa instead of sal-apa-ipa
+        sal_apa_g2p = get_g2p_engine("sal-apa")
+        self.assertEqual(sal_apa_g2p("ac"), list("ats"))
+
+        # but iku-sro goes to iku-sro-ipa, not iku-ipa
+        iku_sro_g2p = get_g2p_engine("iku-sro")
+        self.assertEqual(iku_sro_g2p("akaq"), list("akaq"))
+
     def test_phonemizer_normalization(self):
         moh_g2p = get_g2p_engine("moh")
         self.assertEqual(moh_g2p("\u00e9"), ["\u00e9"])
--- a/everyvoice/text/phonemizer.py
+++ b/everyvoice/text/phonemizer.py
@@ -13,7 +13,7 @@ from everyvoice.config.shared_types import G2PCallable
 DEFAULT_G2P = "DEFAULT_G2P"
 
 
-def make_default_g2p_engines():
+def make_default_g2p_engines() -> dict[str, str | G2PCallable]:
     return {k: DEFAULT_G2P for k in get_arpabet_langs()[0]}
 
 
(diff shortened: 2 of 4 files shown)


everyvoice/tests/test_custom_g2p.py
```diff
@@ -95,6 +95,15 @@ def test_basic_g2p(self):
         with self.assertRaises(NotImplementedError):
             get_g2p_engine("boop")
 
+    def test_unusual_ipa_code(self):
+        # sal-apa goes to sal-ipa instead of sal-apa-ipa
+        sal_apa_g2p = get_g2p_engine("sal-apa")
+        self.assertEqual(sal_apa_g2p("ac"), list("ats"))
+
+        # but iku-sro goes to iku-sro-ipa, not iku-ipa
```
roedoejet, line 103: Hm, why does this not go to `iku-ipa`?

roedoejet: I guess we kind of assumed this *-ipa convention that isn't enforced

everyvoice/text/phonemizer.py
```diff
@@ -32,9 +32,22 @@ def make_default_g2p_engines():
...
+
+        Most languages in the g2p library have a three letter code lll mapped to
+        lll-ipa, but a few do not, e.g., sal-apa -> sal-ipa, oji-syl -> oji-ipa"""
+
+        from g2p.mappings.langs import LANGS_NETWORK
+
+        if lang_id + "-ipa" in LANGS_NETWORK.nodes:
+            return lang_id + "-ipa"
+        else:
+            return lang_id[:3] + "-ipa"
```
roedoejet: hm, what about `lang_id.split('-')[0] + "-ipa"` ?  Do we enforce the initial code will be 3 letters somewhere (I mean, ISO639-3 stipulates this but I don't think we actually test this anywhere.

roedoejet (commented):
Just a couple comments for now. not requesting these changes necessarily, but let's discuss next week.

everyvoice/tests/test_custom_g2p.py
joanise: `g2p show-mappings | grep iku` will tell you that we have iku->iku-equiv->iku-ipa->eng-ipa as the path from syllabics, and the path iku-sro->iku-sri-ipa->iku-sro-ipa->eng-ipa for romanized, and those two paths are just not connected. I don't know why we made the choice, but since we never had an official policy or way to declare "this is the IPA code for language X", whoever wrote the mapping thought that was intuitive to them.

Actually, the git logs tell me that's from back in 2019, with a commit log "first attempt at consolidating langs", so I'm going to guess this might have been an artefact of the merging process. We could change things in g2p, and probably we should add a function to the API that returns the IPA code for any non-IPA code that leads to IPA in a way or another. But my problem would remain that any such solution would be future only, it would not be compatible with older versions of g2p, hence my solution here.

everyvoice/text/phonemizer.py
joanise: https://github.com/NRC-ILT/g2p/pull/489 addresses this question. I was hoping you'd review both PRs together. I'm not attached to any given solution, as long as we pick one and apply it identically to the two PRs.

joanise: But actually, splitting on dash is a nice idea, more forward looking that taking the first 3 characters, so yeah, I'll make this change in both PRs.

everyvoice/model/aligner/wav2vec2aligner
roedoejet, line 1: needs to rebase onto main

roedoejet (approved):
Approving, but please rebase https://github.com/EveryVoiceTTS/wav2vec2aligner/tree/dev.ej/ci-fix-ffmpeg onto https://github.com/EveryVoiceTTS/wav2vec2aligner/tree/main

PR #792, opened 2026-04-29
Use pytest everywhere

joanise:
### PR Goal? 

Make pytest the way we run tests in every circumstance.

At the same time, I removed `silence_c_str{err,out}` because they do nothing when using pytest.

### How to test? 

 - run `pytest`
 - `everyvoice/run_tests.py` does effectively the same, but it still has suites, and `--describe`
 - see CI logs for tests using pytest in that context

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/wav2vec2aligner/pull/29
https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/138

### DIFF
--- a/.github/workflows/matrix-tests.yml
+++ b/.github/workflows/matrix-tests.yml
@@ -34,7 +34,7 @@ jobs:
           pip install -e .[dev]
       - run: pip freeze
       - run: pip list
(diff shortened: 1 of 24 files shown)


roedoejet (approved):
looks good!

Issue #686, opened 2025-06-18
Add StyleTTS2 to EV


roedoejet:
### Description & Motivation

We have done preliminary experiments with StyleTTS2 and would like to add it as a supported e2e model. This is because StyleTTS2 seems to perform well with relatively small amounts of data and is able to handle noisy recordings as well.

### Pitch

This is a very large task that needs to include:

- [x] Turn the code into a library
- [x] Fix imports and style
- [x] Loading HF models for the pre-trained BERT encoder
- [x] Have the auxiliary ASR module integrated as well (or maybe replace with Tim's method)
- [x] Refactoring the model code to Lightning
- [x] Refactor the train and validation loops to Lightning
- [x] Refactor the optimizers to Lightning
- [x] Refactor the dataloader to Lightning
- [x] Store the default text and integrate a type of mapping method for new datasets
- [x] Parameterize everything according to our config strategy
- [ ] Hook up with `everyvoice synthesize`
- [ ] Hook up with `everyvoice demo`
- [ ] Hook up with `everyvoice inspect checkpoint`
- [ ] decoder config matches EveryVoice's 22.05 kHz default, but not StyleTTS2's 24kHz default. I need to think more about what we want to do here.
- [ ] I'm handling the OOD data required by StyleTTS2 kind of sloppily, but improving this will require some thought
- [ ] everyvoice preprocess is currently just using the fs2 preprocess. I kind of think this function should be refactored into the main ev repo
- [ ] the "save-top-k" checkpoints isn't working, it's just saving every epoch
- [ ] only one audio sample seems to be getting logged to tensorboard (should be two or more)
- [ ] fix redundancies in the config (i.e. values that are defined in two places, both the ev config and the styletts2 config
- [ ] The text config is confusing in the StyleTTS2 config because we don't really use the punctuation that gets automatically defined there. We need to consider the symbol mapping requirements for styletts2 more carefully here.
- [ ] everyvoice new-project is very slow now when exporting the yaml
- [ ] when doing everyvoice train text-to-wav config/everyvoice-text-to-wav.yaml on a config file created by EV<=0.4, four extra_forbidden validation errors are produced. This should be easy to catch and output a message explaining the version mismatch and how to fix it.
- [ ] update the readme in styletts2 to reflect our fork's information
- [ ] uses a source-installed version of monotonic align which will block us from publishing on pypi

### Alternatives

[...]

PR #793, opened 2026-05-01
StyleTTS2 Integration

roedoejet:
Work-In-Progress. Feel free to poke around, but not ready for reviews yet.

### PR Goal? 

Merge in the [StyleTTS2](https://github.com/yl4579/StyleTTS2) model into EveryVoice. Use EveryVoice preprocessing, text cleaning, and dataloaders. Build a PyTorch Lightning wrapper around StyleTTS2 and fix the multi-gpu training issue in stage 2 training.

### Fixes? 

https://github.com/EveryVoiceTTS/EveryVoice/issues/686

### Feedback sought? 

This is a really big number of changes on two fronts:

1. a serious overhaul of the original codebase with lightning integration, formatting, cleanup (removing unused files), and packaging (using uv). Because I forked from the original repo, you can see all the change history [here](https://github.com/yl4579/StyleTTS2/compare/main...EveryVoiceTTS:StyleTTS2:main).
2. integration with the new packaged StyleTTS2 repo and EveryVoice. This means adding a Lightning wrapper over the model, integrating EV style configs (which is done using a translation layer - the model still internally uses the old configuration object), integrating EV style data modules, building an EV style CLI so that `everyvoice train text-to-wav` now trains using StyleTTS2. I haven't yet integrated `everyvoice synthesize` or tested `everyvoice demo` among other integrations, but I wanted to get feedback asap once an mvp was finished.

### Priority? 

high - given the size of this, I'd prefer to get this merged in before too much other development takes place.

### Tests added? 

I had to overhaul some of the main repo tests and also added tests in the submodule

### How to test? 

1. Review code to understand the high level changes that took place. There are a lot of integration points with EveryVoice so pointing out things that are missed here would be helpful.
2. Try cloning, configuring a test dataset, and training a text-to-wav model.

### Confidence? 

medium

### Version change? 

yes, minor change.

### Related PRs? 

Since the submodule hasn't been merged in yet, I've just been committing to main there, so there are no other PRs.

### Known Issues

- [ ] uses a source-installed version of monotonic align which will block us from publishing on pypi
- [ ] doesn't integrate with `everyvoice synthesize`
- [ ] haven't checked whether it works with `everyvoice demo`
- [ ] haven't checked whether it works with `everyvoice inspect checkpoint`
- [ ] decoder config matches EveryVoice's 22.05 kHz default, but not StyleTTS2's 24kHz default. I need to think more about what we want to do here.
- [ ] I'm handling the OOD data required by StyleTTS2 kind of sloppily, but improving this will require some thought
- [ ] `everyvoice preprocess` is currently just using the fs2 preprocess. I kind of think this function should be refactored into the main ev repo
- [ ] the "save-top-k" checkpoints isn't working, it's just saving every epoch
- [ ] only one audio sample seems to be getting logged to tensorboard (should be two or more)
- [ ] there are some redundancies in the config (i.e. values that are defined in two places, both the ev config and the styletts2 config
- [ ] The text config is confusing in the StyleTTS2 config because we don't really use the punctuation that gets automatically defined there. We need to consider the symbol mapping requirements for styletts2 more carefully here.
- [ ] `everyvoice new-project` is very slow now when exporting the yaml
- [ ] EJ: when doing `everyvoice train text-to-wav config/everyvoice-text-to-wav.yaml` on a config file created by EV<=0.4, four `extra_forbidden` validation errors are produced. This should be easy to catch and output a message explaining the version mismatch and how to fix it.
- [ ] update the readme in styletts2 to reflect our fork's information
-  [ ] please add any more that you find here...


everyvoice/model/e2e/config.py
```diff
@@ -0,0 +1,6 @@
+from everyvoice.model.e2e.StyleTTS2_lightning.styletts2.ev_config import (  # noqa F401
```
joanise, line 1: This noqa is not necessary

everyvoice/wizard/basic.py
```diff
@@ -221,7 +221,11 @@ def validate(self, response):
         return response in ("yaml", "json")
 
     def effect(self):
-        from everyvoice.model.e2e.config import E2ETrainingConfig, EveryVoiceConfig
+        from everyvoice.model.e2e.config import E2EConfig
+        from everyvoice.model.e2e.StyleTTS2_lightning.styletts2.ev_config import (
+            StyleTTS2ModelConfig,
+            StyleTTS2TrainingConfig,
+        )
```
joanise: Wouldn't it make more sense to expose these in `e2e/config.py` like we make the two fs2 config symbols visible through `feature_prediction/config.py`?

everyvoice/model/e2e/config.py
```diff
@@ -0,0 +1,6 @@
...
+    StyleTTS2Config,
+)
+
```
joanise, line 7: Maybe we should import StyleTTS2ModelConfig and StyleTTS2TrainingConfig here to make them visible to other modules, e.g., `wizard/basic.py`, like we do for other submodule config classes.

everyvoice/cli.py
```diff
@@ -512,15 +517,20 @@ def new_project(
     help="""
     # Train Help
 
+    Please visit http://pathtomodelinfodocs for more information
+
```
joanise, line 521: I would leave a capitalized `TODO` in here so we notice it later...

everyvoice/cli.py
```diff
@@ -512,15 +517,20 @@ def new_project(
...
+        - **text-to-wav** --- this is a model that directly trains text input to waveform/audio output. It is much more heavy duty than training a text-to-spec model. A text-to-spec model will take less than 1 day to train for most datasets, while a text-to-wav model will take over a week for most datasets. For that reason, we strongly recommend training a text-to-wav model first. If you are satisfied with the results, and have extra GPU resources, you can try training a text-to-wav model for improved sentence-level audio quality.
+
     """,
 )
 
 train_group.command(
     name="text-to-spec",
-    short_help="Train your Text-to-Spec model",
+    short_help="Train your Text-to-Spec (FastSpeech2) model",
```
joanise, line 533: Thank you, that's really helpful to name the models like this. That confused me a lot when I first started working EV.

pyproject.toml
```diff
@@ -63,26 +63,35 @@ dependencies = [
   "nltk==3.9.3",
   "packaging>=22.0",
   "pandas~=2.0",
-  "protobuf~=4.25",                # https://github.com/EveryVoiceTTS/EveryVoice/issues/387
+  "protobuf~=4.25", # https://github.com/EveryVoiceTTS/EveryVoice/issues/387
```
joanise, line 66: I bet we don't need to lock protobuf back anymore, maybe more versions have been releases since, they probably fixed the problem we were seeing...

pyproject.toml
```diff
@@ -63,26 +63,35 @@ dependencies = [
...
   "pydantic[email]>=2.4.2,<2.8.0",
   "pympi-ling",
   "pyworld-prebuilt==0.3.4.4",
   "PyYAML>=6.0",
   "readalongs>=1.2.0",
   "questionary>=2.0.0",
-  "setuptools<80",  # needed for tensorboard us to 2.20.0, should no longer be needed with next version
+  "setuptools<80", # needed for tensorboard us to 2.20.0, should no longer be needed with next version
```
joanise, line 73: Can we ditch this yet?

joanise (commented):
Mostly finicky comments while reading through the code changes here to get familiar with this all.
Need to start playing with the models next.
I added one more improvement to your list in the PR description.

everyvoice/wizard/basic.py
```diff
@@ -472,14 +476,31 @@ def effect(self):
...
-                vocoder=vocoder_config,
-                training=E2ETrainingConfig(
+                model=StyleTTS2ModelConfig(multispeaker=multispeaker),
+                preprocessing=preprocessing_config,  # TODO: should we use the default 24kHz sampling rate that StyleTTS2 usually uses? if so, the hop size, upsample rates etc will also need to change
+                pretrained=StyleTTS2PretrainedConfig(
+                    f0_path=pretrained_folder / "jdc/bst.t7",
+                    asr_config=pretrained_folder / "asr/config.yml",
+                    asr_path=pretrained_folder / "asr/epoch_00080.pth",
+                    plbert_dir=pretrained_folder / "plbert",
+                ),
```
joanise, line 499: I understand what you're doing here and why, but there is an unfortunate side effect of storing the full path to the everyvoice software, which will often encode the username, in the config and therefore in the model checkpoints we'll generate.
If we can come up with a different way to point to it, that would be preferable.

everyvoice/wizard/basic.py
```diff
@@ -472,14 +476,31 @@ def effect(self):
...
+                training=StyleTTS2TrainingConfig(
+                    ood_data=st_dir
```
joanise, line 501: ditto this

everyvoice/wizard/basic.py
roedoejet: yes, good point. I'll refactor this.

pyproject.toml
roedoejet: Doesn't seem like it: https://github.com/tensorflow/tensorboard/issues/7003#issuecomment-4239111787

everyvoice/model/e2e/config.py
roedoejet: it is actually, at least with the recent refactors

joanise (commented):
Oops, I forgot to post these two comments.

PR #794, opened 2026-05-01
Refactor a lot of test assertions to pytest style

joanise:
### PR Goal? 

Moving further along in the pytest conversion, this PR takes advantage of the nicer assertions and fixtures that pytest provides.

The idea going forward is to always write pytest-style tests, with `assert <expr>` everywhere, and no longer use `TestCase` at all. This PR does not go 100% there, but it moves significantly in that direction. For the submodules, however, it's a 100% completed conversion.

### Feedback sought? 

sanity checking I guess. It's a long diff, but it only touches tests files.

### Priority? 

normal

### How to test? 

- look at CI results
- look at coverage, it should not drop since the refactored tests are still the same
- manually: compare running `pytest` in the parent branch and running `pytest` in this one.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/wav2vec2aligner/pull/30
https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/139


joanise:
Note: I will clean up the commit history before merging.

roedoejet (approved):
lgtm - thanks @joanise !

PR #795, opened 2026-05-05
refactor(ci): move the speed test into a separate script

joanise:
### PR Goal? 

Factor out the code to check the time it takes to run `everyvoice -h` to make sure it stays fast

### Fixes? 

legibility, plus the .sh script is lintable

### Priority? 

normal

### Tests added? 

no

### How to test? 

Mostly look at the output of the "Make sure the CLI stays fast" step in CI, and the PR comment. They should stay as before.

You can also run `./profile-help-ci.sh` but better pipe the output into less, it's optimized for CI logs, not interactive use. This will leave around two text files, `importtime.txt` and `import-message.txt`. That's on purpose, and they're now .gitignore'd. You can use them for analysis of what's wrong in your PR, too, if your PR is slowing things down.

### Confidence? 

pretty good

### Version change? 

nope

### Related PRs? 

nope

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,6 @@ jobs:
           coverage xml
       - run: cd everyvoice && coverage report
       - name: Check for logs_and_checkpoints/ and preprocessed/
-        id: no-extra-directories
         run: |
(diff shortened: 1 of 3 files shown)


roedoejet (approved):
thanks @joanise

PR #796, opened 2026-05-05
Dev.ej/fail fast help

joanise:


joanise:
was just for testing #795 to see what happened in the case of failures. This PR's tests failed as desired, which means a pass for 795. 😄

Issue #385, opened 2024-04-11
New command to check compatibility between a filelist and a model or config as well as update the text config based on the filelist


roedoejet:
there should be a way to check that a given filelist is compatible (i.e. can be used for training/inference) with a text config.

The command should check to see if the filelist has the appropriate text representation (characters/phones) and whether they are tokenized (character_tokens/phone_tokens). It should tokenize if tokens aren't already available. Then it should make a set of all the symbols in the filelist and check that they are all contained in the text config. 

The `update` version would then add the missing symbols to the text config. 

This would be useful for the situation where someone wants to see if a given filelist is compatible with their model's current configuration. The `update` function would invalidate the model, but would be helpful for the situation where someone:

- runs the wizard for a language that does not have g2p
- then later adds g2p
- then wants to preprocess their phones but their text configuration does not contain the phone list

I imagine being able to check with either a text config or a model (where we just read the text config stored in the model).

Something like `everyvoice check-filelist <filelist> <model|text-config>` or `everyvoice update-text-config <filelist> <text-config>`. The update function should have a warning that it will invalidate any models trained with the previous config.

PR #799, opened 2026-05-06
Check and update text config for new text

joanise:
### PR Goal? 

Add a new subcommand `check-text-config` to validate if some new text is compatible with a model or config in terms of the character inventory it requires.

### Fixes? 

Does parts of #385

### Feedback sought? 

Careful review: does this meet the requirements, does it work correctly, and what additional work will still be required?

If this works but is incomplete, I'll want to merge this and make future PRs for other features. Of course, anything that isn't right I'll want to fix before merging.

~Also, I'm in the process of splicing out commits that you will find here that are not germane to this PR, to make reviewing easier. You can already start testing though.~ Unrelated changes were now spliced out, this is ready to review. The remaining semi-related changes included here are too closely tied to this PR to splice out.

### Priority? 

pretty high

### Tests added? 

yup!

### How to test? 

`everyvoice check-text-config --config config/everyvoice-shared-text.yaml --text-file some_text.txt --language und`

and other variants as documented in `everyvoice check-text-config -h`

### Confidence? 

medium

### Version change? 

no

### Related PRs? 

no

### DIFF
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -1333,12 +1333,16 @@ def check_text_config(
     config: Annotated[
         Optional[Path],
         typer_file_option(
-            help="path to text config, i.e., everyvoice-shared-text.yaml"
+            "--config",
+            "-c",
+            help="path to text config, i.e., everyvoice-shared-text.yaml",
         ),
     ] = None,
     model: Annotated[
         Optional[Path],
-        typer_file_option(help="path to a model whose text config will be used"),
+        typer_file_option(
+            "--model", "-m", help="path to a model whose text config will be used"
+        ),
     ] = None,
     text_file: Annotated[
         Optional[Path],
@@ -1351,8 +1355,10 @@ def check_text_config(
     language: Annotated[
         Optional[str],
         typer.Option(
+            "--language",
+            "-l",
             help="language id, required with --text-file, or for a psv file without a language column. "
-            + "Declaring the language is always required, because text normalization can be language specific, and g2p is always language specific."
+            + "Declaring the language is always required, because text normalization can be language specific, and g2p is always language specific.",
         ),
     ] = None,
 ):
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -1308,7 +1308,8 @@ def open_text_or_psv_file(
         if language is None:
             raise typer.BadParameter("--language is required with --text-file.")
         records = [{"characters": line, "language": language} for line in text_lines]
-    elif psv_file:
+    else:
+        assert psv_file
         records = generic_psv_filelist_reader(psv_file)
         if "language" not in records[0]:
             if language is None:
@@ -1317,8 +1318,6 @@ def open_text_or_psv_file(
                 )
             for record in records:
                 record["language"] = language
-    else:
-        assert False
     return records
 
 
@@ -1329,9 +1328,15 @@ def get_text_config_from_config_or_model(config: Optional[Path], model: Optional
 
     if config:
         text_config: TextConfig = TextConfig.load_config_from_path(config)
-    elif model:
+    else:
+        assert model
         with spinner("Loading model"):
-            checkpoint = load_checkpoint(model)
+            try:
+                checkpoint = load_checkpoint(model)
+            except Exception as e:
+                raise typer.BadParameter(
+                    f"Model/checkpoint '{model}' does not appear to be valid.\nError from loader: {e}"
+                )
         # print("Looking for text config")
         model_config = checkpoint["hyper_parameters"]["config"]
         if "text" in model_config:
@@ -1341,7 +1346,7 @@ def get_text_config_from_config_or_model(config: Optional[Path], model: Optional
         else:
             # Models without text config, e.g., a HiFiGan Vocoder, are not accepted here
             raise typer.BadParameter(
-                f"Model/checkpoint {model} does not have an embedded text configuration."
+                f"Model/checkpoint '{model}' does not have an embedded text configuration."
             )
     return text_config
 
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -538,35 +538,7 @@ command(
     """,
 )(preprocess_styletts2)
 
-app.add_typer(
-    preprocess_group,
-    name="preprocess",
-    short_help="Preprocess your data",
-)
-
-
-# Add check_data to root
-command(
-    app,
-    name="check-data",
-    short_help="Check your data for outliers or any anomalies",
-    help="""
-    # Check Data Help
-
-    This command will check all of your data to help you find anomalies and outliers.
-
-    To check your data, make sure you've run preprocessing first (everyvoice preprocess --help).
-    Then you need to briefly and partially train a text-to-spec model. We recommend 100-1000 steps to start.
-
-    Then, with your partially trained model you can run the data checker:
-    \n\n
-    **everyvoice check-data config/everyvoice-text-to-spec.yaml logs_and_checkpoints/FeaturePredictionExperiment/base/checkpoints/last.ckpt**
-    \n\n
-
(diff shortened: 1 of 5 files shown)


everyvoice/preprocessor/preprocessor.py
```diff
@@ -824,23 +831,27 @@ def process_text(
...
-            )
-            assert isinstance(tokens, list)
-            phone_tokens = tokens
-        # calculate pfs
-        if phone_tokens and use_pfs:
-            pfs = text_processor.calculate_phonological_features(
-                text_processor.token_sequence_to_text_sequence(phone_tokens),
-                apply_punctuation_rules=True,
-            )
+        if (
```
roedoejet, line 834: thanks for cleaning this all up

everyvoice/tests/test_cli.py
```diff
@@ -46,13 +46,19 @@ def major_minor(version):
 
 COMMANDS = [
     "new-project",
+    "preprocess",
     "train",
     "synthesize",
-    "preprocess",
+    "demo",
+    "check-data",
```
roedoejet: I think maybe `check-data` and `check-text-config` can be merged into `everyvoice check data` and `everyvoice check text-config`

everyvoice/cli.py
```diff
@@ -85,8 +86,27 @@
     TEXT_TO_WAV_CONFIG_FILENAME_PREFIX,
 )
 
+
+# For the main `everyvoice` command, TyperGroupOrderAsDeclared doesn't work because
+# single commands gets listed first, followed by sub-command groups. There are so many
+# top-level commands and command groups now that it gets quite confusing, so let's order
+# them logically for the main ones, and alphabetically for the rest.
+class MainCommandOrder(typer.core.TyperGroup):
```
roedoejet, line 89: awesome! Yes, I've been meaning to do this - thanks!

everyvoice/cli.py
```diff
@@ -1254,5 +1273,169 @@ def g2p(
...
+                raise typer.BadParameter(
+                    "--language is required for a psv file without a language column."
+                )
+            for record in records:
+                record["language"] = language
+    return records
+
+
+def get_text_config_from_config_or_model(config: Optional[Path], model: Optional[Path]):
+    """Helper for chec_text_config: load a TextConfig from a config file or model file"""
```
roedoejet:
```suggestion
    """Helper for check_text_config: load a TextConfig from a config file or model file"""
```

everyvoice/cli.py
```diff
@@ -1254,5 +1273,169 @@ def g2p(
...
+            try:
+                checkpoint = load_checkpoint(model)
+            except Exception as e:
+                raise typer.BadParameter(
+                    f"Model/checkpoint '{model}' does not appear to be valid.\nError from loader: {e}"
+                )
+        # print("Looking for text config")
+        model_config = checkpoint["hyper_parameters"]["config"]
+        if "text" in model_config:
+            # Question: FS2 models have text config, do any others have it?
```
roedoejet: Yes, StyleTTS2 has it as well, but not HiFiGAN

everyvoice/cli.py
```diff
@@ -1254,5 +1273,169 @@ def g2p(
...
+            # For other models that have it, are they in the same place in the metadata?
```
roedoejet: Yep: https://github.com/EveryVoiceTTS/StyleTTS2/blob/0ab9c316917a9ca1b42f06e94202990ad3fc7004/styletts2/ev_config/__init__.py#L561

everyvoice/cli.py
```diff
@@ -1254,5 +1273,169 @@ def g2p(
...
+        ),
+    ] = None,
+):
+    """
+    Inspect a text configuration for compatiblity with an input file
+
+    Test processing input_file against the text configuration provided, or the text
+    configuration found in model, and report any incompatibilities.
+
+    Required options: one of --config and --model, as well as one of --text-file and --psv-file.
```
roedoejet: Hm, could we not have --file and then try to inspect it? not a big deal, but this tripped me up on first-try.

everyvoice/cli.py
```diff
@@ -1254,5 +1273,169 @@ def g2p(
...
+    records = open_text_or_psv_file(text_file, psv_file, language)
+
+    # Expensive imports are deferred so we fail fast where we can
+    with spinner("Loading software"):
+        from everyvoice.config.text_config import TextConfig  # noqa F401
+        from everyvoice.preprocessor.preprocessor import Preprocessor
+        from everyvoice.text.text_processor import TextProcessor
+        from everyvoice.text.utils import guess_graphemes_in_text
+
+    text_config = get_text_config_from_config_or_model(config, model)
```
roedoejet: If I pass a text-to-spec config for example, I get a somewhat cryptic error. Can we handle this by catching ValidationErrors and trying to load text-to-spec/spec-to-wav etc configs and if one of them passes, then raising that as an error? Alternatively, the text config is actually available in the text-to-spec config, so it would ideally be best if we look for `text` in the config and try and load that if it exists as well.

Image

roedoejet (changes requested):
This is so awesome @joanise. I'm really excited to use this! I've put a couple suggestions/recommendations on here that I'd like to see. Some or maybe all you can address before merging, otherwise it's OK to merge and make issues.

Thanks!

everyvoice/tests/test_cli.py
joanise: good idea

everyvoice/cli.py
joanise: I don't see the text config in our styletts2 checkpoint, so I'm not sure how I'll be able to support that model type...

```
everyvoice checkpoint inspect logs_and_checkpoints/E2E-Experiment/base/last.ckpt
```

is giving me  no `model_info`, and no `hyper_parameters`, or no `text` field.

everyvoice/cli.py
joanise: Right, that looks like what I need, but again I don't see it in the checkpoint...

I can see, though, that we could support

    everyvoice check test-config --config config/everyvoice-text-to-wav.yaml ...

everyvoice/cli.py
joanise: We could, but I'm not a super fan. Maybe worth separating into its own issue.

The logic would be a bit messy:
 - if `|` occurs in the first line, we would presumably assume .psv, but how would we support a literal `|` in the first piece of text?
 - a single column "psv" (with `characters` as the first line) should be supported -- no `|` on the first line
 - if the first utterance is literally `characters`, or `phones`, that's ambiguous again

everyvoice/cli.py
joanise: Directly supporting the text-to-spec config is a good idea.

everyvoice/cli.py
joanise: In the wizard we take the time to ask clarifying questions from the user, but I'd rather not write that logic up again. Every user will trip on this, but I think the error message is clear enough.

everyvoice/tests/test_cli.py
joanise: done

everyvoice/cli.py
joanise: Figured it out. I had an old model, in a new model I found the text config, although not quite in the same place as in an FS2 checkpoint.

everyvoice/cli.py
joanise: never mind, figured it out

everyvoice/cli.py
joanise: Deferred to issue #844

everyvoice/cli.py
roedoejet: fair enough!

roedoejet (approved):
thanks for the improvements @joanise - this looks great!

PR #800, opened 2026-05-07
ci: update setup-miniconda for node 24 bump

joanise:
### PR Goal? 

[conda-incubator/setup-miniconda](https://github.com/conda-incubator/setup-miniconda) finally released a patch fully using node 24, let's bump.

### Editorial comment

We only use conda in CI for MacOS in matrix testing, where I have not succeeded in using non-conda solutions to install sox. But I'd love to move off of Conda completely here too, looking for a contributor to do it, without loosing too much efficiency. 

Expected results: replacing conda by alternative sox and ffmpeg installers and uv for the python packages would most likely speed up the workflow, but my expectations are often wrong for what's fast or slow in CI... So dang unpredictable!

### Feedback sought?

Rubber stamping for the PR, a maybe volunteer for addressing my editorial comment.

### DIFF
--- a/.github/workflows/matrix-tests.yml
+++ b/.github/workflows/matrix-tests.yml
@@ -51,7 +51,7 @@ jobs:
         with:
           submodules: recursive
       - name: Set up conda
-        uses: conda-incubator/setup-miniconda@fc2d68f6413eb2d87b895e92f8584b5b94a10167 # v3.3.0
+        uses: conda-incubator/setup-miniconda@8ee1f361103df19b6f8c8655fd3967a8ecb162d5 # v4.0.1
         with:
           python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
on Mac CI you could install with brew too but this won't be any faster than conda I don't think

joanise:
> on Mac CI you could install with brew too but this won't be any faster than conda I don't think

I think I tried that before. I should have kept better notes, but I vaguely recall it didn't work. Either it didn't end up properly visible on the path, or maybe libsox.so wasn't it. I could try again, since we so longer need libsox.

joanise:
Nice, it does work now, and it's actually faster with brew install than miniconda. Closing this PR and submitting another to remove conda altogether instead.

PR #801, opened 2026-05-07
ci: on MacOS, use brew install ffmpeg and sox; remove conda

joanise:
### PR Goal? 

This PR replaces #800: instead of updating setup-miniconda, remove it! In the past, `brew install sox` did not work, with `libsox.so` not being found, but since we don't need libsox.so anymore, that problem is moot.

So now our matrix tests for MacOS just use the standard setup-python, with `brew install ffmpeg sox` for those dependencies, rather than conda. Brew turns out to be faster.

### Feedback sought? 

sanity check

### Priority? 

normal

### Tests added? 

n/a

### How to test? 

Look at the CI logs for the matrix-tests workflow: https://github.com/EveryVoiceTTS/EveryVoice/actions/runs/25517910389

### Confidence? 

high

### DIFF
--- a/.github/workflows/matrix-tests.yml
+++ b/.github/workflows/matrix-tests.yml
@@ -15,6 +15,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.11", "3.12"]
+      fail-fast: false
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
lgtm

PR #803, opened 2026-05-12
clean up work around removing pytest as a prod dependency

### LINKED ISSUE #802: Remove pytest from production requirements
### Bug description

I'm accidentally made pytest a prod requirements in either #792 or #794, this needs to be undone.

We could either:
 - fix `everyvoice/cli.py` so it does not indirectly import `pytest`, while still preserving the hidden `everyvoice test` command
 - or, since I don't believe anyone ever uses `everyvoice test`, and just running `pytest` is actually shorter, I think we should remove `everyvoice test` from the CLI instead.

### How to reproduce the bug

Checkout `origin/main` and run `everyvoice -h` when pytest is not installed:
 - either do a fresh venv and only install `uv pip install -e .`
 - or from an existing venv run `uv pip uninstall pytest`

### Error messages and logs

```
$ everyvoice -h
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\joanise\sandboxes\EveryVoice\.venv\Scripts\everyvoice.exe\__main__.py", line 4, in <module>
  File "C:\Users\joanise\sandboxes\EveryVoice\everyvoice\cli.py", line 67, in <module>
    from everyvoice.run_tests import SUITE_NAMES, run_tests
  File "C:\Users\joanise\sandboxes\EveryVoice\everyvoice\run_tests.py", line 12, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

### Environment

_No response_

### More info

_No response_

joanise:
### PR Goal? 

tidying up stuff...
 - `pytest` should not be a prod dependency
 - in CI, run the CLI speed test before installing test extras, so we now it at least starts without them
 - docs: I felt our docs about how to get ffmpeg from their site needed a little clarifying, I made this change a while ago but it's been sitting there, not worth its own PR...
 - when reviewing the StyleTTS2 PR, I stumbled on `update-schemas`, not remembering where the schemas were, so I decided to improve the error message, and while I was in there I figured it would simple to have it consider an up-to-date schema file OK even if it already exists. Only out-of-sync schema files that would get overwritten should be considered an error.

### Fixes? 

Fixes #802

### Feedback sought? 

sanity check

### Priority? 

normal

### Tests added? 

yes

### How to test? 

Run `everyvoice update-schemas` and enjoy the friendly thumbs-up like output.

Run
```
echo asdf >> everyvoice/.schema/everyvoice-shared-text-0.4.json
everyvoice update-schemas
```
and enjoy a helpful message on what to do about the out of date file.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

Yeah, this conflicts with #793, but I rebased it to `dev.ej/styletts2-rebased` to pre-resolve the conflict for you.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,9 +41,21 @@ jobs:
           # Install a specific version of uv.
           version: "0.11.8"
       - name: Install the project
-        run: uv sync --locked --all-extras --dev
+        run: uv sync --locked
       - run: uv pip freeze
       - run: uv pip list
+      - name: Make sure the CLI stays fast
+        run: |
+          source .venv/bin/activate
+          ./profile-help-ci.sh "${{ github.event.pull_request.head.sha }}"
+      - name: Report help speed in PR
+        if: github.event_name == 'pull_request'
+        uses: mshick/add-pr-comment@7c1a3a3e5a072270dc21c0639df39ca11e860b2b # v3.5.0
+        with:
+          preformatted: true
+          message-path: import-message.txt
+      - name: Install test dependencies
+        run: uv sync --locked --extra test
       - name: Run tests
         run: |
           cd everyvoice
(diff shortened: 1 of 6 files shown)


roedoejet (approved):
looks good to me @joanise

PR #804, opened 2026-05-15
[StyleTTS2] feat: add demo and inference integration basics

roedoejet:
### PR Goal? 

I previously connected the `everyvoice train` functionality with StyleTTS2. This PR integrates StyleTTS2 with `everyvoice demo`, `everyvoice synthesize`, and `everyvoice checkpoint inspect` commands.

### Fixes? 

Part of https://github.com/EveryVoiceTTS/EveryVoice/issues/686

### Feedback sought? 

Testing, but also sanity. I think there are actually quite a few places where we tied ourselves a bit too closely to FS2 and its architecture. I think I need some space to be able to tell how to refactor, but any insight is helpful.

I'm mostly looking for high-level analysis about whether the approach to combine repos in this way is reasonable.

### Priority? 

high

### Tests added? 

none so far

### How to test? 

try running `everyvoice synthesize`, `everyvoice demo`, and `everyvoice checkpoint inspect`. note, that I don't think this will work on the model you *just* trained. I'm not adding backwards compatibility support for that, although the hooks are in place for us to be able to handle this in the future in the same way as FS2

### Confidence? 

medium

### Version change? 

n/a, already bumped to 0.5.0

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/142
https://github.com/EveryVoiceTTS/StyleTTS2/pull/4

### DIFF
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -67,7 +67,6 @@ from everyvoice.model.vocoder.HiFiGAN_iSTFT_lightning.hfgl.cli import (
     synthesize as synthesize_hfg,
 )
 from everyvoice.model.vocoder.HiFiGAN_iSTFT_lightning.hfgl.cli import train as train_hfg
-from everyvoice.utils import spinner
 from everyvoice.wizard import (
     PREPROCESSING_CONFIG_FILENAME_PREFIX,
     SPEC_TO_WAV_CONFIG_FILENAME_PREFIX,
@@ -790,6 +789,8 @@ def _run_styletts2_demo(
     print(f"  - Share:          {share}")
     print(f"  - Server Name:    {server_name}")
 
+    from everyvoice.utils import spinner
+
     with spinner("Loading software"):
         from everyvoice.demo.app import create_demo_app_styletts2
 
@@ -865,6 +866,8 @@ def _run_fs2_demo(
     print(f"  - Share:          {share}")
     print(f"  - Server Name:    {server_name}")
     ui_config_json = _load_fs2_ui_config(ui_config_file)
+
+    from everyvoice.utils import spinner
 
     with spinner("Loading software"):
         from everyvoice.demo.app import create_demo_app
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -659,317 +659,118 @@ AllowedDemoOutputFormats = Enum(  # type: ignore
     [("all", "all")] + [(i.name, i.value) for i in SynthesizeOutputFormats],
 )
 
-
-# Add the demo commands
-demo_group = typer.Typer(
-    pretty_exceptions_show_locals=False,
-    no_args_is_help=True,
-    context_settings={"help_option_names": ["-h", "--help"]},
-    rich_markup_mode="markdown",
-    cls=TyperGroupOrderAsDeclared,
-    help="""
-    # Demo Help
-
-        - **text-to-spec** --- Launch an interactive Gradio demo for a two-stage (FastSpeech2 + HiFiGAN) model.
-
-        - **text-to-wav** --- Launch an interactive Gradio demo for an end-to-end (StyleTTS2) model.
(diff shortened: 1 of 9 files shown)


joanise:
To resolve the conflict, keep `SCHEMAS_TO_OUTPUT:` and its preceding comment just before  `def update_schemas`.

everyvoice/demo/app.py
```diff
@@ -474,6 +474,244 @@ def load_model_from_checkpoint(
...
+                )
+                btn = gr.Button("Synthesize")
+            with gr.Column():
+                out_audio = gr.Audio(format="wav", label="Output Audio")
+
+        btn.click(synthesize_fn, inputs=inputs, outputs=[out_audio])
+    return demo
+
+
+def create_demo_app_styletts2(
```
joanise, line 644: Is there any code at all that is shared between the styletts2 demo and the fs2+hfgl one? As I scan, it seems like it's completely separate code, and in that case I would think that this should be placed in `styletts2/cli/demo.py` instead of here.

That would also facilitate adding it to the styletts2 CLI as `styletts2 demo`, actually.

everyvoice/cli.py
```diff
@@ -669,9 +680,29 @@ def test(suite: TestSuites = typer.Argument("dev")):  # pragma: no cover
...
+
+        - **text-to-spec** --- Launch an interactive Gradio demo for a two-stage (FastSpeech2 + HiFiGAN) model.
+
+        - **text-to-wav** --- Launch an interactive Gradio demo for an end-to-end (StyleTTS2) model.
+    """,
+)
+
+
+@demo_group.command(
+    name="text-to-spec",
```
joanise: Hum, this name won't be intuitive for me: this demo variant does text->spec->wav, while the other does text->wav: both generate wav from text.

I just discussed it with Sam, and he came up with `text-to-spec-to-wav` for this vs `text-to-wav` for the other. It's longer, but with auto complete you don't have to type it all. Sam also found that we can create aliases if we don't want to have to always use the long names.

E.g., 
```
@demo_group.command(name="text-to-spec-to-wav" short_help="...")
@demo_group.command(name="t2s2w", hidden=True)
def demo...()
```
would define an alias command `everyvoice demo t2s2w` if we want to have a shorter version available. The alias could even use the model short names, maybe `everyvoice demo fs2-hfgl` vs `everyvoice demo styletts2` could be our aliases for the two variants?

joanise (commented):
Not really tested yet, but I'm done for today. Generally good but see suggestions below.

everyvoice/cli.py
roedoejet: hm, yes, I totally see what you mean. I do like this better than what I have, but I'm also thinking that maybe we can skirt through this whole issue by reading the metadata in the checkpoint since EV declares which type of model it is. That way, we could just run `everyvoice demo path/to/ckpt` and let the command figure out which one of these commands to run. That would be cleaner from a UX point of view, because there isn't really any reason the user needs to know exactly what type of model it is when they want to demo it. If you agree with this, I can implement and push.

joanise: Absolutely, that would be great! And we have all the logic needed for inspect, why not use it here.

Obviously, it should be an error to provide a vocoder and a styletts, and a different error to have fs2 without a vocoder, but that's pretty simple logic.

The other advantage of your suggestion: current demo startup commands keep working unchanged.

roedoejet: OK - I'll give this a go

everyvoice/demo/app.py
roedoejet: I agree that this would be good, but I might leave it for another PR since it requires re-factoring code that isn't part of this PR feature. For now, I'll leave this demo code where the other demo code lives, but I'll also make an issue to refactor it in the way you suggest.

joanise (approved):

PR #805, opened 2026-05-15
[StyleTTS2] Fix training checkpointing

roedoejet:
Previously, training would log every epoch. Now, it mimics FS2 - that is, log the last epoch as well as the top 5 best ckpts in terms of Validation Mel Spectrogram loss. It will also fix the audio-logging issue where only one audio file was being logged to tensorboard. I also added text logging. 

To test, try running everyvoice train and inspect the tensorboard logs.

https://github.com/EveryVoiceTTS/StyleTTS2/pull/5


joanise (approved):
lgtm

PR #807, opened 2026-05-19
[StyleTTS2] load pretrained checkpoint from HuggingFace

roedoejet:
### PR Goal? 

Loading pretrained models from within the codebase is messy since we need to handle the paths, and the model checkpoints bloat the package/our git history.

### Fixes? 

An element of #686 

### Feedback sought? 

Does this look reasonable? I think the "local override" option is still not going to be handled well since it will put a potentially absolute path in the config.

### Priority? 

high

### Tests added? 

None

### How to test? 

if you train a model with this branch, you should see the HF models downloading automatically

### Confidence? 

medium

### Version change? 

n/a

### Related PRs? 

https://github.com/EveryVoiceTTS/StyleTTS2/pull/6

Once accepted, we'll need to clean the git history of any *.t7 or *.pth files

### DIFF
--- a/everyvoice/.schema/everyvoice-text-to-wav-0.5.json
+++ b/everyvoice/.schema/everyvoice-text-to-wav-0.5.json
@@ -445,6 +445,61 @@
       "title": "Punctuation",
       "type": "object"
     },
+    "StyleTTS2ASRConfig": {
+      "$schema": "http://json-schema.org/draft-07/schema#",
+      "additionalProperties": false,
+      "description": "Source for the ASR text-aligner checkpoint.",
+      "properties": {
+        "repo_id": {
+          "default": "everyvoice/styletts2-asr-aligner",
+          "description": "HuggingFace repo ID for the ASR text-aligner.",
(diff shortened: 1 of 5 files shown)


joanise:
the new-project wizard still expected the pre-trained files to exist

joanise:
The concept is good here.

For the local override, we could put a warning in the field's documentation that it gets saved in the models. Also, we could support `$HOME/path/to/...` fairly easily with `os.path.expandvars()` and `os.path.expanduser()` which could get called after pydantic validation, or maybe in post-validation, as long as the input string is what gets saved in the config.

roedoejet:
> The concept is good here.
> 
> For the local override, we could put a warning in the field's documentation that it gets saved in the models. Also, we could support `$HOME/path/to/...` fairly easily with `os.path.expandvars()` and `os.path.expanduser()` which could get called after pydantic validation, or maybe in post-validation, as long as the input string is what gets saved in the config.

good idea. I've added this.

joanise (approved):

PR #812, opened 2026-05-27
Replace monotonic-align from source by our published ilt-monotonic-align

joanise:
### PR Goal? 

Get ilt-monotonic-align from PyPI instaed of github, so we can publish the next EveryVoice version.

This is a draft to review the publication process of ilt-monotonic-align, which is currently only published on test.pypi.org.
Please test with this PR and if we're happy with it, then I'll actually publish to regular PyPI.

Source code for our ilt-monotonic-align fork: https://github.com/EveryVoiceTTS/monotonic_align

### Fixes? 

Fixes https://github.com/EveryVoiceTTS/StyleTTS2/issues/8

### Feedback sought? 

Please test and make sure it works for you.

Also please review 
https://github.com/EveryVoiceTTS/monotonic_align/compare/c6e5e6cb19...783173dd7d
which is the diff between the repo we forked and what I have done, specifically for how I declare Copyright and maintainer, and general sanity checking. The intent at this point was to do the minimal changes required to be able to publish, so ignore "would-be-nice" elements, just review for actual problems.

### Priority? 

before we publish 0.5.0

### Tests added? 

yes

### How to test? 

By default, pip or uv won't fetch packages from testpypi, you need to add `--index-strategy unsafe-best-match --extra-index-url https://test.pypi.org/simple` to make that happen.

The lock file `uv.lock` file is already updated to point to testpypi for this one package, so just
```
uv sync --locked
```
will work.
(Once we publish ilt-monotonic-align to PyPI, I'll update the lock file again accordingly.)

Otherwise, you can do
```
uv pip install --index-strategy unsafe-best-match --extra-index-url https://test.pypi.org/simple -e .
```

Then, launch training of a styletts2 model and make sure it gets past the point where it need monotonic-align. @roedoejet when is that anyway?

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/StyleTTS2/pull/10

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,7 @@ dependencies = [
   "transformers>=4.30",
   "einops-exts",
   "torch==2.7.1",
-  "monotonic-align",
+  "ilt-monotonic-align",
   "torchaudio==2.7.1",
   "torchinfo==1.8.0",
   "tqdm>=4.66.0",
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
tested with a newly built env, can confirm it works. thanks so much @joanise !

PR #814, opened 2026-05-28
ci: alternative way to install ffmpeg on failure in matrix too

joanise:
### PR Goal? 

CI fails too often when we cannot install ffmpeg. I installed this alternative thing in tests.yaml before, now extending it to matrix testing too.

### Priority? 

low

### Confidence? 

high

### Version change? 

no


roedoejet (approved):
good idea - thanks @joanise

PR #815, opened 2026-06-04
fix(tests): adjust test to pass with recent typer

joanise:


roedoejet (approved):
Ugh, that's annoying! nice fix though!

PR #816, opened 2026-06-11
Support StyleTTS2 OOD data requirements 

roedoejet:
### PR Goal? 

Figure out how to handle OOD data properly. This is data that is used for the WavLM based loss used in StyleTTS2. It intentionally uses text (from the same language) to synthesize that is unseen during training, to make the synthesizer more robust to out of domain utterances. This PR:

- moves the default English data to HuggingFace: https://huggingface.co/datasets/everyvoice/StyleTTS2-English-OOD
- adds ood to the preprocessor (exports a per-language psv file)
- adds ood steps to the wizard
- allows the user to provide a HF repo link, a local file, or use the validation set

### Fixes? 

Fixes part of https://github.com/EveryVoiceTTS/EveryVoice/issues/686

### Feedback sought? 

This is a big one, so mostly looking for feedback on the UX in the wizard for handling OOD data. Is it explained OK? How intuitive is it? Try configuring + preprocessing + training if you can.

### Priority? 

high. this is one of the last things for StyleTTS2, and probably the second-to-last major change.

### Tests added? 

I added a few.

### How to test? 

Configure a new project, add some OOD data, preprocess, train.

### Confidence? 

medium-high

### Version change? 

already changed for StyleTTS2

### Related PRs? 

https://github.com/EveryVoiceTTS/StyleTTS2/pull/13

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -152,6 +152,7 @@ git submodule foreach 'prek install'
 git submodule foreach 'gitlint install-hook'
 ```
 
+
 ## Pull request recommendations
 
 ### TL;DR
@@ -208,3 +209,46 @@ This command can be quite helpful in doing this inspection:
 git submodule foreach "git log --oneline --graph --decorate HEAD origin/main | head"
 ```
 carefully look at the output for each submodule, making sure you agree with where HEAD points.
+
+## Publishing Instructions
+
+To publish a new version of the project, follow these steps:
+
+1. **Determine the Version Bump**
+   Decide whether your changes constitute a:
+   - **Major** version bump (breaking changes),
+   - **Minor** version bump (new features, backward-compatible, any change to the schema), or
+   - **Patch** version bump (bug fixes, small changes).
+
+2. **Update Version Files**
+   - Update the `everyvoice._version` file to reflect the new version.
+   - Keep all `submodule._version` files in sync with this version, **except** for the `wav2vec2aligner` submodule (which can be installed separately).
+   - Keep the `everyvoice` dependency in all `submodule/pyproject.toml` files in sync with the everyvoice Major.minor version, except in `wav2vec2aligner`.
+   - Commit the resulting changes, including all submodules.
+
+3. **Update the Documentation**
+   - Make sure the documentation reflects the current state of the code.
+   - Look for references to the current or most recent version and update them if necessary.
+
+4. **Update Schema (for Major/Minor bumps)**
+   If you bumped a **major** or **minor** version:
+   - Run `everyvoice update-schema`. You may need to delete existing schema files if you get an error message, but you should only do so if you are sure that those schema files have not already been published. I.e. we might already have schema files related to an alpha release - those can be overwritten, but we should never change published schema files.
+   - Commit the resulting changes.
+
+5. **Open a Pull Request**
+   - Create a PR with your changes.
+   - Wait for tests to pass and for the PR to be merged into `main`.
+
+6. **Tag the Release**
+   After merging:
+   ```bash
+   git tag -a -m vX.Y.Z vX.Y.Z
+   git push 'vX.Y.Z'
+   ```
+
+7. **Update SchemaStore (for Major/Minor bumps)**
+    Once the CI has built and released your version, if you bumped a major or minor version:
+
+    Submit a PR to [SchemaStore](https://github.com/SchemaStore/schemastore) to update the schema reference.
+
+    The only file you need to change is: `src/api/json/catalog.json`
--- a/docs/guides/custom.md
+++ b/docs/guides/custom.md
@@ -39,6 +39,12 @@ You can also use the "festival" format which is like this (example from [Sinhala
 
 In this format, there are corresponding wav files labelled sin_2241_0329430812.wav etc..
 
+### StyleTTS2 OOD Data
+
+One of the models supported by EveryVoice, named StyleTTS2, uses 'out-of-distribution' (OOD) data during training. Unlike standard TTS data, OOD text does **not** need corresponding audio. StyleTTS2 uses a pretrained WavLM discriminator that only judges whether generated audio sounds real or fake, so no audio is required for that loss. For best results, use text from the same language but outside your training set (hence "out of distribution").
+
+The configuration wizard (Step 4) will prompt you for this text. You can point to a file on your computer or on HuggingFace, but it must have a `characters` or `phones` header line. The wizard will also offer to use your validation data as OOD data, but this is not recommended since it can artificially lower your validation loss.
+
 ## Step 3: Install EveryVoice
 
 Head over to the [installation documentation](../install.md) and install EveryVoice
--- a/docs/guides/custom.md
+++ b/docs/guides/custom.md
@@ -62,7 +62,7 @@ cd your_everyvoice_project
 
 ## Step 5: Choose a Model
 
-EveryVoice supports two model architectures, and the right choice depends on what you want to prioritize. The Configuration Wizard 🧙 will have already asked you which model you want to train — this section explains the difference so you know what you're getting into.
+EveryVoice supports two model architectures, and the right choice depends on what you want to prioritize. This section explains the difference so you know what you're getting into.
 
 **[FastSpeech2](./fastspeech2.md)** is a two-stage pipeline: a feature prediction network converts text to mel spectrograms, and then a separate vocoder converts those spectrograms to audio. It trains relatively quickly (usually less than a day for most datasets on a single GPU), and produces good, consistent speech. If this is your first time training a TTS model with EveryVoice, or if you want reliable results without a lot of fuss, this is a good place to start. Even if you plan on training a StyleTTS2 model, this is a good low-cost option for training a model with your data. If your system isn't intelligible with FastSpeech2, it certainly won't work with StyleTTS2, and the solution is likely to be found in fixing issues in your data. If your FastSpeech2 system sounds pretty good, and you want to take it to the next level, try StyleTTS2.
 
--- a/docs/guides/custom.md
+++ b/docs/guides/custom.md
@@ -58,118 +58,17 @@ cd your_everyvoice_project
 ```
 
 !!! important
-After you run the Configuration Wizard 🧙, please inspect your text configuration `config/{{ config_filename('text') }}` to make sure everything looks right. That is, if some unexpected symbols show up, please inspect your data (if you remove symbols from the configuration here, they will be ignored during training). Sometimes characters that are treated as punctuation by default will need to be removed from the punctuation list if they are treated as non-punctuation in your language.
+    After you run the Configuration Wizard 🧙, please inspect your text configuration `config/{{ config_filename('text') }}` to make sure everything looks right. That is, if some unexpected symbols show up, please inspect your data (if you remove symbols from the configuration here, they will be ignored during training). Sometimes characters that are treated as punctuation by default will need to be removed from the punctuation list if they are treated as non-punctuation in your language.
 
-## Step 5: Run the Preprocessor
+## Step 5: Choose a Model
 
-Your models need to do a number of preprocessing steps in order to prepare for training. To preprocess everything you need, run the following:
+EveryVoice supports two model architectures, and the right choice depends on what you want to prioritize. The Configuration Wizard 🧙 will have already asked you which model you want to train — this section explains the difference so you know what you're getting into.
 
-```bash
-everyvoice preprocess config/{{ config_filename('text-to-spec') }}
-```
+**[FastSpeech2](./fastspeech2.md)** is a two-stage pipeline: a feature prediction network converts text to mel spectrograms, and then a separate vocoder converts those spectrograms to audio. It trains relatively quickly (usually less than a day for most datasets on a single GPU), and produces good, consistent speech. If this is your first time training a TTS model with EveryVoice, or if you want reliable results without a lot of fuss, this is a good place to start. Even if you plan on training a StyleTTS2 model, this is a good low-cost option for training a model with your data. If your system isn't intelligible with FastSpeech2, it certainly won't work with StyleTTS2, and the solution is likely to be found in fixing issues in your data. If your FastSpeech2 system sounds pretty good, and you want to take it to the next level, try StyleTTS2.
 
-## Step 6: Select a Vocoder
+**[StyleTTS2](./styletts2.md)** is an end-to-end model that goes directly from text to audio in a single pass, using a diffusion process to model speaking style and other unlabelled variation (i.e. recording environment, minor dialect differences). At synthesis time it takes a short reference audio clip and uses it to match the style. Training is split into two stages and it takes a lot longer than FastSpeech2 (over a week with multiple GPUs on most datasets), but can produce particularly natural-sounding speech, especially for expressive or varied speaking styles.
 
-You do not need to train your own vocoder.
-EveryVoice is compatible out-of-the-box with the UNIVERSAL_V1 HiFiGAN checkpoint from [the official HiFiGAN implementation](https://github.com/jik876/hifi-gan?tab=readme-ov-file#pretrained-model), which is very good quality. You can find the EveryVoice-compatible version of this checkpoint [here](https://drive.google.com/drive/folders/1ya0U4K2d26DoJamg96cEynMJ1w1Tm8nU?usp=sharing).
+If you're not sure which to pick, start with FastSpeech2.
 
-You can download the checkpoint by following the link above, or you can download it using the command line with [gdown](https://pypi.org/project/gdown/). First ensure that you have _gdown_ installed; you can install it with `pip install gdown`. Then to download the checkpoint you can run:
-
-```bash
-gdown https://drive.google.com/uc?id=1-iarZV2hTeociQjTX7l2WShuXalROGQf
-```
-
-Using a pre-trained vocoder is recommended, and the above checkpoint should work well even for new languages after [finetuning](./finetune.md).
-
-### Train your own Vocoder
-
-You might want to train your own vocoder, but this takes a long time (up to 2 weeks on a single GPU), uses a lot of electricity, and unless you know what you are doing, you are unlikely to improve upon the publicly available models discussed above, even for a new language. So we do not recommend it. You are almost always better off just using the pre-trained vocoder and then [finetuning](./finetune.md) on the predictions from your feature prediction network. If you really do want to train your own vocoder though, you can run the following command:
-
-```bash
-everyvoice train spec-to-wav config/{{ config_filename('spec-to-wav') }}
-```
-
-By default, we run our training with PyTorch Lightning's "auto" strategy. But, if you are on a machine where you know the hardware, you can specify it like:
-
-```bash
-everyvoice train spec-to-wav config/{{ config_filename('spec-to-wav') }} -d 1 -a gpu
-```
-
-Which would use the GPU accelerator (`-a gpu`) and specify 1 device/chip (`-d 1`).
-
-## Step 7: Train your Feature Prediction Network
-
-To generate audio when you train your feature prediction network, you need to add your vocoder checkpoint to the `config/{{ config_filename('text-to-spec') }}`
-
-At the bottom of that file you'll find a key called `vocoder_path`. Add the absolute path to your trained vocoder (here it would be `/path/to/test/logs_and_checkpoints/VocoderExperiment/base/checkpoints/last.ckpt` where `/path/to` would be the actual path to it on your computer.)
-
-Once you've replaced the `vocoder_path` key, you can train your feature prediction network:
-
-```bash
-everyvoice train text-to-spec config/{{ config_filename('text-to-spec') }}
-```
-
-!!! tip
-While your model is training, you can use TensorBoard to view the logs which will show information about the progress of training and display spectrogram images. If you have provided a `vocoder_path` key, then you will also be able to hear audio in the logs. To use TensorBoard, make sure that your conda environment is activated and run `tensorboard --logdir path/to/logs_and_checkpoints`. Then your logs will be viewable at [http://localhost:6006](http://localhost:6006).
-
-## Step 8 (optional): Finetune your Vocoder
-
-When you have finished training your Feature Prediction Network, we recommend [finetuning](./finetune.md) your vocoder. This step is optional, but it will help get rid of metallic artefacts that are often present if you don't finetune your vocoder. Note, it will likely not help with any mispronounciations. If you notice these types of errors, it is likely due to issues with the training data (e.g. too much variation in pronunciation or recording quality in the dataset, or discrepencies between the recording and transcription.)
-
-## Step 9: Synthesize Speech in Your Language!
-
-#### Command Line
-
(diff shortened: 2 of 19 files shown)


docs/guides/custom.md
```diff
@@ -58,118 +58,17 @@ cd your_everyvoice_project
...
+**[StyleTTS2](./styletts2.md)** is an end-to-end model that goes directly from text to audio in a single pass, using a diffusion process to model speaking style and other unlabelled variation (i.e. recording environment, minor dialect differences). At synthesis time it takes a short reference audio clip and uses it to match the style. Training is split into two stages and it takes a lot longer than FastSpeech2 (over a week with multiple GPUs on most datasets), but can produce particularly natural-sounding speech, especially for expressive or varied speaking styles.
 
-Please refer to `everyvoice evaluate --help` for more information.
+If you're not sure which to pick, start with FastSpeech2.
 
-!!! note
-Automatic evaluation can be helpful, but please take the reported numbers with a grain of salt. They are not always reliable, and do not always correlate well with human judgements.
+- Continue with **[FastSpeech2 →](./fastspeech2.md)**
```
joanise, line 83: Great idea to split the documentation this way with sub-pages for the two variants.

docs/guides/fastspeech2.md
```diff
@@ -0,0 +1,117 @@
+# Training a FastSpeech2 Model
+
+This page picks up from [Step 5: Choose a Model](./custom.md#step-5-choose-a-model) in the main guide. If you haven't completed Steps 1–4 yet, start there first.
```
joanise: Since you reset the numbering from 1 here, `Steps 1-4` is ambiguous. Maybe "If you haven't completed steps 1-4 described there yet"? I don't like it, if you have a better way to lift the ambiguity that would be nice.

docs/guides/fastspeech2.md
```diff
@@ -0,0 +1,117 @@
...
+Your model needs to do a number of preprocessing steps in order to prepare for training. To preprocess everything you need, run the following:
+
+```bash
+everyvoice preprocess text-to-spec config/{{ config_filename('text-to-spec') }}
+```
+
+## Step 2: Select a Vocoder
+
+You do not need to train your own vocoder.
+EveryVoice is compatible out-of-the-box with the UNIVERSAL_V1 HiFiGAN checkpoint from [the official HiFiGAN implementation](https://github.com/jik876/hifi-gan?tab=readme-ov-file#pretrained-model), which is very good quality. You can find the EveryVoice-compatible version of this checkpoint [here](https://drive.google.com/drive/folders/1ya0U4K2d26DoJamg96cEynMJ1w1Tm8nU?usp=sharing).
```
joanise: Hum, if we need the EV-compatible version, then it's not really "compatible out-of-the-box", is it?

docs/guides/fastspeech2.md
```diff
@@ -0,0 +1,117 @@
...
+everyvoice train spec-to-wav config/{{ config_filename('spec-to-wav') }} -d 1 -a gpu
+```
+
+Which would use the GPU accelerator (`-a gpu`) and specify 1 device/chip (`-d 1`).
+
+## Step 3: Train your Feature Prediction Network
+
+To generate audio when you train your feature prediction network, you need to add your vocoder checkpoint to the `config/{{ config_filename('text-to-spec') }}`
+
+At the bottom of that file you'll find a key called `vocoder_path`. Add the absolute path to your trained vocoder (here it would be `/path/to/test/logs_and_checkpoints/VocoderExperiment/base/checkpoints/last.ckpt` where `/path/to` would be the actual path to it on your computer.)
```
joanise: Hum, I never noticed this before. When I test, I always leave that config as generated by the wizard. Is that a bad thing? For testing I suppose it does not really matter, but for real training, does this step matter?

joanise: Or is it just the audio in the tensorboard I'm forfeiting?

joanise: It might be worth hinting here at why this matters, even if it's repeated in the tensorboard tip.

docs/guides/fastspeech2.md
```diff
@@ -0,0 +1,117 @@
...
+```bash
+everyvoice train text-to-spec config/{{ config_filename('text-to-spec') }}
+```
+
+!!! tip
+    While your model is training, you can use TensorBoard to view the logs which will show information about the progress of training and display spectrogram images. If you have provided a `vocoder_path` key, then you will also be able to hear audio in the logs. To use TensorBoard, make sure that your conda environment is activated and run `tensorboard --logdir path/to/logs_and_checkpoints`. Then your logs will be viewable at [http://localhost:6006](http://localhost:6006).
+
+## Step 4 (optional): Finetune your Vocoder
+
+When you have finished training your Feature Prediction Network, we recommend [finetuning](./finetune.md) your vocoder. This step is optional, but it will help get rid of metallic artefacts that are often present if you don't finetune your vocoder. Note, it will likely not help with any mispronounciations. If you notice these types of errors, it is likely due to issues with the training data (e.g. too much variation in pronunciation or recording quality in the dataset, or discrepencies between the recording and transcription.)
```
joanise: typo: discrepencies  -> discrepancies

docs/guides/styletts2.md
```diff
@@ -0,0 +1,79 @@
+# Training a StyleTTS2 Model
+
+This page picks up from [Step 5: Choose a Model](./custom.md#step-5-choose-a-model) in the main guide. If you haven't completed Steps 1–4 yet, start there first.
```
joanise: same comment as for the same line in the fs2 page

joanise (commented):
Just my comments so far, I'll keep reviewing next week.

I know some of my comments are on text that you didn't just write, but rather only moved, but I noticed things we could improve anyway.

And I know, so far only on docs, nothing on the code yet. Next week...

everyvoice/cli.py
```diff
@@ -472,25 +475,88 @@ def new_project(
...
 
+preprocess_group.command(
+    name="text-to-wav",
+    no_args_is_help=True,
+    short_help="Preprocess data for text-to-wav (StyleTTS2) training",
+    help=f"""Preprocess data for a StyleTTS2 text-to-wav model.
+
+    **everyvoice preprocess text-to-wav config/{TEXT_TO_WAV_CONFIG_FILENAME_PREFIX}.yaml**
+
+    To also preprocess an out-of-domain (OOD) text file:
```
joanise: In the styletts2 repo (or PR?) this was "out-of-distribution", which I think is the correct term to use here, no?

docs/guides/styletts2.md
```diff
@@ -0,0 +1,79 @@
+# Training a StyleTTS2 Model
```
joanise, line 1: I think this file needs a subsection on OOD, and it should spell out that OOD means (I think out of distribution). We need to at least hint at what that is, and why you may or may not care about it. It's a bit of a mysterious concept, so it's important to at least mention it and give some advice on what decent options are if you don't want to work hard at it.

everyvoice/wizard/basic.py
```diff
@@ -206,6 +209,264 @@ def effect(self):
         )
 
 
+class OODDataStep(Step):
+    """Ask user for the OOD data source for a specific language."""
+
+    REVERSIBLE = True
+    choices = (
+        "validation: use the validation split (warning: pollutes train/validation separation). If you don't intend to use StyleTTS2, or if you don't have any more data, please select this option.",
```
joanise: Maybe

```
... (warning: pollutes train/validation separation for StyleTTS2 models). ...
```

everyvoice/wizard/basic.py
```diff
@@ -206,6 +209,264 @@ def effect(self):
...
+    )
+
+    def __init__(self, lang: str, **kwargs):
+        super().__init__(**kwargs)
+        self.lang = lang
+
+    def prompt(self):
+        rich_print(
+            Panel(
+                "OOD (out-of-domain) texts are used by StyleTTS2 for calculating WavLM discriminator loss. "
```
joanise: Hum, you seem to have decided to switch to "out of domain" rather that "out of distribution", as evidenced here and in a few other places. I'm confused. I thought some text in the same language just not in the training data would do. In my mind, that's not actually out of domain. It might be out of the specific training data distribution, but it's still pretty dang close to being in-domain!

everyvoice/wizard/basic.py
```diff
@@ -206,6 +209,264 @@ def effect(self):
...
+        else:  # validation
+            self.saved_state = {
+                "ood_raw_data": deepcopy(self.state.get("ood_raw_data"))
+            }
+            self.state["ood_raw_data"][self.lang] = {"source_type": "validation"}
+            rich_print(
+                Panel(
+                    "[yellow]Warning: using the validation split as OOD data will pollute your "
+                    "train/validation separation and may make your validation scores unreliable."
+                    "Unless you are training your model for research purposes, this is probably OK."
```
joanise, line 273: When I'm planning to train an fs2 model, it might be confusing to have to answer this question.

everyvoice/wizard/basic.py
```diff
@@ -206,6 +209,264 @@ def effect(self):
...
+        except RepositoryNotFoundError:
+            rich_print(
+                f"[yellow]Could not access repository '[bold]{self.repo_id}[/bold]'. "
+                "Proceeding anyway.[/yellow]"
+            )
+        except Exception:
+            rich_print(
+                "[yellow]Could not reach HuggingFace Hub to verify the file. "
+                "Proceeding anyway.[/yellow]"
+            )
```
joanise: Strange these exception cases are returning True. Do we have a fallback to fetch this later on? I guess in the Wizard, we're just doing early caching, are we?

everyvoice/wizard/basic.py
```diff
@@ -206,6 +209,264 @@ def effect(self):
...
+            with open(Path(response).expanduser(), encoding="utf-8") as fh:
+                first_line = fh.readline()
+        except OSError as e:
+            rich_print(f"[red]Could not read the file: {e}[/red]")
+            return False
+        self._detected_repr = _detect_ood_text_representation(first_line)
+        if self._detected_repr is None:
+            rich_print(
+                "[red]The file does not have a pipe-separated header with a "
+                "'characters' or 'phones' column. Your text file should be a pipe separated file (basename|characters) and must have either a characters or phones column.[/red]"
```
joanise, line 327: `basename` should be optional here, if you're only going to read the characters or the phones column, and if we're going to suggest using other text in the language that might not be recorded.

It's confusing here because saying `(basename|characters)` suggests you're asking me for more text with accompanying audio files.

everyvoice/wizard/basic.py
joanise: Maybe add something like "We will download the file later" after "Proceeding anyway"?

everyvoice/wizard/basic.py
```diff
@@ -230,297 +491,340 @@ def effect(self):
...
-            wavs_dir = Path(dataset_state[StepNames.wavs_dir_step]).expanduser()
-            if not wavs_dir.is_absolute():
-                if not output_path.is_absolute():
-                    for _ in config_dir.parts:
-                        wavs_dir = Path("..") / wavs_dir
-                else:
-                    wavs_dir = Path.cwd() / wavs_dir
-            new_filelist_path = (
-                Path("..") / f"{dataset_name}-filelist.psv"
+        with spinner("Preparing your configuration files"):
```
joanise, line 495: The spinner should have been kept in a prior PR, as it is it's going to be very hard to review the changes in this function...

everyvoice/cli.py
```diff
@@ -472,25 +475,88 @@ def new_project(
...
+
+        - **text-to-wav** --- preprocess data for a StyleTTS2 (text-to-wav) model.
+            Only audio and text steps are run; pitch, energy, and spectrogram
+            features are computed on-the-fly during training.
+    """,
+)
+
+
+@preprocess_group.callback(invoke_without_command=True)
+def _preprocess_deprecated(ctx: typer.Context):
```
joanise: `no_args_is_help=True` makes this unnecessary: we already get the help message when called without a command. I suspect as things are, this function can never do anything.

I could see the value if we were able to intercept `everyvoice preprocess  config/everyvoice-text-to-spec.yaml` and give the suggestion here, but that just gives me `No such command 'config/everyvoice-text-to-spec.yaml'.` and I'm not sure how you could possibly fix this.

everyvoice/tests/test_cli.py
```diff
@@ -448,6 +453,18 @@ def test_preprocessing_with_wrong_config(self):
...
+        """'everyvoice preprocess' without a subcommand shows available subcommands."""
+        result = self.runner.invoke(app, ["preprocess"])
+        # no_args_is_help=True causes the help text to list both subcommands
+        assert "text-to-spec" in flatten_log(result.output)
+        assert "text-to-wav" in flatten_log(result.output)
+
+    def test_preprocess_text_to_wav_help(self):
+        """'everyvoice preprocess text-to-wav --help' should exit cleanly."""
+        result = self.runner.invoke(app, ["preprocess", "text-to-wav", "--help"])
+        assert result.exit_code == 0
```
joanise: This assertion is unstable. See #815.

everyvoice/tests/test_wizard.py
```diff
@@ -1595,7 +1680,11 @@ def test_multilingual_multispeaker_true_config(self) -> None:
                 ),
                 RecursiveAnswers(
                     patch_menu_prompt(0),
-                    children_answers=[RecursiveAnswers(patch_menu_prompt(0))],
+                    children_answers=[
+                        RecursiveAnswers(patch_menu_prompt(0)),  # OODDataStep[eng]
```
joanise, line 1775: Somewhere in `text_wizard.py` (not likely right here, I just picked a random instance of OODDataStep) I would have liked to see hf and local exercised through these recursive answers.

docs/guides/fastspeech2.md
roedoejet: good point, no it's not

docs/guides/fastspeech2.md
roedoejet: good point, I moved all this to the tip

joanise (commented):
Comments from ~today~ yesterday -- I managed to play with the wizard and review most of the code in this PR -- I have not yet started to review the styletts2 PR.

joanise:
FYI, I just fixed the failing licensecheck in CI.

everyvoice/tests/test_wizard.py
```diff
@@ -484,6 +484,78 @@ def test_no_data_to_save(self):
...
+            [basic.OODDataStep(lang="und", name="OOD Data Step [und]")],
+        )
+        tour.state["ood_raw_data"] = {}
+        step = tour.steps[0]
+        with patch_menu_prompt(0):  # 0 = validation
+            step.run()
+        assert step.children == ()
+        assert tour.state["ood_raw_data"]["und"] == {"source_type": "validation"}
+
+    def test_ood_local_source(self):
```
joanise, line 501: it would be good to add error cases in here, with a file not found, a file without PSV headers, a file without the needed column.

See `test_wrong_fileformat_psv` for an example patching multiple answers in just one block:
```
with patch_menu_prompt((1, 2, 3, 0), multi=True) as stdout:
            format_step.run()
```
This provides three answer I know will fail validation (1, 2, and 3 in this case) and then a final answer I know will pass (0 here) for which I then make assertions that confirm it's the one I expected to pass whose effect is apparent in the tour state, and also assertions of expected messages in the logs:
```
output = flatten_log(stdout.getvalue())
self.assertRegex(output, r"does not look like a .*'tsv'")
self.assertRegex(output, r"does not look like a .*'csv'")
assert "is not in the festival format" in output
assert format_step.completed
```

That's the lightest weight solution I found to exercising validation errors without having to write a lot of testing code, although of course you'd still have to create the invalid input files. Although `os.devnull` is a fine choice for an empty file you don't have to create...

joanise: Ditto the validation for the other steps you wrote, it would be worth devising test cases for at least some of their error situations, although thank you for stubbing huggingface_hub and not letting unit tests actually talk to HF.

everyvoice/base_cli/helpers.py
```diff
@@ -133,8 +133,6 @@ def preprocess_base_command(
 
     config = load_config_base_command(model_config, config_args, config_file)
     preprocessor = Preprocessor(config)
-    if isinstance(config, StyleTTS2Config):
-        steps = ["audio", "text"]
     if (
         (isinstance(config, FastSpeech2Config) or isinstance(config, StyleTTS2Config))
```
joanise: minor detail: this is equivalent to `isinstance(config, (FastSpeech2Config, StyleTTS2Config))`

everyvoice/cli.py
roedoejet: yes! thanks, a good reason to spell-it-out elsewhere too

everyvoice/wizard/basic.py
roedoejet: yes, I'll do a find-replace for this. it is out-of-(training)distribution but should be in the same domain

everyvoice/cli.py
roedoejet: riight, ok, so let's delete then?

everyvoice/wizard/basic.py
roedoejet: yea, but I'd kind of rather have them configure everything from the get-go instead of only select FS2 and have to re-do this step?

everyvoice/wizard/basic.py
roedoejet: Actually I'll remove this and let it fail.

everyvoice/wizard/basic.py
roedoejet: ya sorry!

everyvoice/tests/test_wizard.py
roedoejet: are the `test_ood_local_source` and `test_ood_hf_source` tests not enough?

joanise: Ambivalent. We can let coverage analysis answer this question.

everyvoice/wizard/basic.py
```diff
@@ -206,6 +209,258 @@ def effect(self):
         )
 
 
+class OODDataStep(Step):
+    """Ask user for the out-of-distribution (OOD) data source for a specific language."""
+
+    REVERSIBLE = True
+    choices = (
+        "validation: use the validation split ... (warning: pollutes train/validation separation). ... If you don't intend to use StyleTTS2, or if you don't have any more data, please select this option.",
```
joanise: Just tested the wizard, this text gets lost: the choices don't get line wrapped, they get truncated. I think you should keep just "validation: use the validation split" here, and move everything else into the prompt, which does get wrapped properly in its panel.

everyvoice/wizard/basic.py
```diff
@@ -206,6 +209,258 @@ def effect(self):
...
+        else:  # validation
+            self.saved_state = {
+                "ood_raw_data": deepcopy(self.state.get("ood_raw_data"))
+            }
+            self.state["ood_raw_data"][self.lang] = {"source_type": "validation"}
+            rich_print(
+                Panel(
+                    "[yellow]Warning: using the validation split as OOD data will pollute your "
+                    "train/validation separation and may make your validation scores unreliable."
+                    "Unless you are training your model for research purposes, this is probably OK."
```
joanise: You need a space after "unreliable.", right now it prints like `unreliable.Unless`.

Contributing.md
```diff
@@ -208,3 +209,46 @@ This command can be quite helpful in doing this inspection:
...
+1. **Determine the Version Bump**
+   Decide whether your changes constitute a:
+   - **Major** version bump (breaking changes),
+   - **Minor** version bump (new features, backward-compatible, any change to the schema), or
+   - **Patch** version bump (bug fixes, small changes).
+
+2. **Update Version Files**
+   - Update the `everyvoice._version` file to reflect the new version.
+   - Keep all `submodule._version` files in sync with this version, **except** for the `wav2vec2aligner` submodule (which can be installed separately).
+   - Keep the `everyvoice` dependency in all `submodule/pyproject.toml` files in sync with the everyvoice Major.minor version, except in `wav2vec2aligner`.
```
joanise, line 226: fs2 and hfgl don't declare a version requirement, they just say `everyvoice`. Only StyleTTS2 is specific. We never try to maintain any compatibility between EV and the submodules beyond the specific sha1 committed in EV, I wonder if we need to be stricter in all the pyproject.toml files. Then again, if we don't publish the submodules to PyPI, it doesn't really matter. I'm only looking at this because you moved it in this PR, but we don't really need to fix it given our publication process.

joanise: Further comments on this, specific to styletts2, in https://github.com/EveryVoiceTTS/StyleTTS2/issues/14

docs/guides/custom.md
```diff
@@ -39,6 +39,12 @@ You can also use the "festival" format which is like this (example from [Sinhala
 
 In this format, there are corresponding wav files labelled sin_2241_0329430812.wav etc..
 
+### StyleTTS2 OOD Data
+
+One of the models supported by EveryVoice, named StyleTTS2, uses 'out-of-distribution' (OOD) data during training. Unlike standard TTS data, OOD text does **not** need corresponding audio. StyleTTS2 uses a pretrained WavLM discriminator that only judges whether generated audio sounds real or fake, so no audio is required for that loss. For best results, use text from the same language but outside your training set (hence "out of distribution").
+
+The configuration wizard (Step 4) will prompt you for this text. You can point to a file on your computer or on HuggingFace, but it must have a `characters` or `phones` header line. The wizard will also offer to use your validation data as OOD data, but this is not recommended since it can artificially lower your validation loss.
```
joanise:
```suggestion
The configuration wizard (Step 4) will prompt you for this text. You can point to a file on your computer or on HuggingFace, but it must be a pipe-separated-value (PSV) file, with a header line, and containing a `characters` or a `phones` column. At minimum, a plain-text file with `characters` on the first line will do, as long as it can be parsed as a one-column PSV file.

The wizard will also offer to use your validation data as OOD data, but this is not recommended when training StyleTTS2 since it can artificially lower your validation loss.

Note that while the wizard will always ask you for OOD data, it is only used for training StyleTTS2 models, not FastSpeech2 models.
```

everyvoice/wizard/basic.py
```diff
@@ -206,6 +209,258 @@ def effect(self):
...
+    REVERSIBLE = True
+
+    def __init__(self, lang: str, **kwargs):
+        super().__init__(**kwargs)
+        self.lang = lang
+        self._detected_repr = None
+
+    def prompt(self):
+        return questionary.path(
+            f"Path to OOD plain-text file for language '{self.lang}. Your text file should be a pipe separated file (e.g., basename|characters) and must have either a 'characters' or 'phones' column': ",
```
joanise:
```suggestion
            f"Path to OOD plain-text file for language '{self.lang}. Your text file must be a pipe separated file (e.g., basename|characters) and must have either a 'characters' or a 'phones' column': ",
```

joanise (approved):
Great work, this is almost ready to merge.

There are a few more changes I would recommend, but I don't need to review again before you merge, so clicking Approve.

PR #828, opened 2026-06-18
Wizard: allow the user to recover from a stale resume file

joanise:
### PR Goal? 

It's poor UX that when you have a resume file from a previous version of EV, you have to edit it by hand because a new question got added, especially if that question is near the end like in the OOD case.

This PR prompts you to decide if you want to continue from the discrepancy point or not.

Also, while testing this, I noticed the spinner at the end of the wizard made the congrats message get printed wrong -- it just had to be printed after the spinner was done instead of inside the spinner scope.

### Priority? 

normal

### Tests added? 

yes

### How to test? 

in the regression suite, the OOD PR makes the resume files stale. With this PR they work anyway until the OOD question and then you can pick from there by hand.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

no


roedoejet (approved):
this is great @joanise - thank you!

PR #830, opened 2026-06-19
refactor: factor out common app configuration settings

joanise:
### PR Goal? 

Factor out common app configuration settings we want to use everywhere:
 - typer.Typer constructor: use `typer.Typer(**default_typer_args, help="...")` everywhere for multiple default settings
 - @app.command decorator: use `@command(app)` everywhere to set `no_args_is_help=True` by default

Also, while I was touching the command groups, I noticed that `everyvoice preprocess` has two of our three `preprocess()` functions but was missing HFG's, so I added it to complete the command group.

### Fixes? 

Make the code DRYer.

### Feedback sought? 

Sanity checking

### Priority? 

normal

### Tests added? 

Most of this code is already all exercised in existing tests.

### How to test? 

The only change you'll notice in practice is that where you used to get

> Error: Missing argument 'MODEL_PATH'

now you'll get the actual help message. E.g., compare `everyvoice synthesize text-to-wav` in main and in this PR.

We already had the no-args-is-help in a lot of places, now it's systematic, except where a command has no mandatory option or argument.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

Yup, in all four submodules!

https://github.com/EveryVoiceTTS/HiFiGAN_iSTFT_lightning/pull/64
https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/148
https://github.com/EveryVoiceTTS/wav2vec2aligner/pull/31
https://github.com/EveryVoiceTTS/StyleTTS2/pull/16

### DIFF
--- a/everyvoice/base_cli/__init__.py
+++ b/everyvoice/base_cli/__init__.py
@@ -20,6 +20,15 @@ default_typer_args = {
 }
 
 
-def command(app: typer.Typer, no_args_is_help=True, **argv):
-    """Wrapper around app.command reversing the default value on no_args_is_help"""
-    return app.command(no_args_is_help=no_args_is_help, **argv)
+def command(app: typer.Typer, no_args_is_help=True, **kwargs):
+    """Wrapper around app.command setting no_args_is_help=True by default
+
+    Usage: replace `@app.command(...)` by:
+
+        @command(app, ...)
+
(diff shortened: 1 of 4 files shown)


roedoejet (approved):
solid DRY refactor - thanks @joanise !

PR #835, opened 2026-07-02
fix: make regression test suite Windows compatible

joanise:
### PR Goal? 

Fixes for Windows:
 - use junctions instead of symlinks, so we don't end up with copies instead
 - make sure the output is utf8

### Feedback sought? 

rubber stamping

### Priority? 

low

### Tests added? 

n/a

### How to test? 

```
cd everyvoice/tests/regression
mkdir delme
cd delme
../prep-datasets.sh
```

and see that on all OSes, you get a `wavs` symlink in each subdirectory created, not a copy.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

no


roedoejet (approved):
I don't have a Windows machine to test this on, but it seems reasonable to me.

PR #836, opened 2026-07-13
Dev.ej/refactor tests

joanise:
### PR Goal? 

Do more unittest -> pytest conversions. It's not complete, but it's work I did a couple weeks ago and then put aside. I'd like to merge it and get back to other things for now.

The most significant improvement with this work is the stubbed models now factored out and defined in `conftest.py`: the are available to all unit tests, and with `scope="session"` no matter how many tests use them, only one dummy model is created and shared among all tests.

Such fixtures is where pytest shines, and the real motivation behind moving away from the `unittest` framework.

### Fixes? 

With this PR, only one stubbed vocoder and one stubbed fs2 model are created per invocation of `pytest`.

### Feedback sought? 

Mostly sanity checking, but I realize the diff is very long. All tests still pass, and I've been careful to maintain that with each step of refactoring, so a full review is not really necessary.

### Priority? 

medium

### How to test? 

Run `pytest`

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/149

### DIFF
--- a/everyvoice/conftest.py
+++ b/everyvoice/conftest.py
@@ -1,3 +1,8 @@
+"""
+Fixtures shared by all unit tests.
+Places in everyvoice instead of everyvoice/tests so that submodules also see it.
+"""
+
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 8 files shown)


roedoejet (approved):
It's kind of just magic that you can pass the name of the session fixture into any function. cool stuff, thanks @joanise !

joanise:
> It's kind of just magic that you can pass the name of the session fixture into any function. cool stuff

Yeah, it feels like magic at first, but once you're used to it, it's quite powerful. I discovered `conftest.py` recently, where you can declare fixtures and they're available in files that don't import it--more magic. And then there are the built-in fixtures you can use without declaring anywhere at all! https://docs.pytest.org/en/stable/reference/fixtures.html

Fixtures have their own set of scoping rules, related to but not identical to usual scoping rules. As I'm learning it all, I like it more and more, but it does require learning a new way to do stuff.

PR #837, opened 2026-07-15
Dev.ej/fix regression

joanise:
### PR Goal? 

Make the regression test suite work again.

As we changed the code, we let this get stale. We should be more diligent making sure CLI changes are reflected in the regression tests.

Note: the apparently unrelated `norecursedirs = [ "build", "dist", ".git", "regress-*", "*venv*", ".mypy_cache" ]` in `pyproject.toml` is essential here: when you run pytest after having created regression tests, the default discovery rules that pytest uses will search through the hundred thousand files that the regression suite creates, and we have to block that to keep things sane. The `testpaths` parameter also declared there only has an effect when you run from the EveryVoice sandbox root, but the `norecursedirs` setting is always taken into account, so excluding large directories using the latter is more important.

### Priority? 

normal

### How to test? 

Run the regression suite and see that it works again

### Confidence? 

high

### DIFF
--- a/everyvoice/tests/regression/go.sh
+++ b/everyvoice/tests/regression/go.sh
@@ -43,7 +43,7 @@ for DIR in regress-*; do
     popd
 done
 
-coverage run -p -m everyvoice test 2>&1
+coverage run -p -m pytest ../../../.. 2>&1
 
 JOB_COUNT=$(find . -maxdepth 1 -name regress-\* | wc -l)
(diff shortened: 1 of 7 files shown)


roedoejet (approved):
I'll try to remember this, but I wonder if we could have some way of reminding/triggering a run here when version bumps happen or something? Seems we've been stale for a couple versions.

PR #838, opened 2026-07-16
build(deps): force replace pydub by pydub_ng

joanise:
### PR Goal? 

`pydub` and `pydub-ng` cannot be installed at the same time: they provide the same files. `gradio` wants the former, `readalongs` wants the latter. Turns out `uv` has an override mechanism to handle that!

https://github.com/ReadAlongs/Studio/pull/289 for more details about pydub vs pydub-ng

### Fixes? 

gets rid of warnings from unmaintained pydub

### How to test? 

See that CI's "test" job, "Run tests" step no longer has warnings `DeprecationWarning: invalid escape sequence '\('` about `pydub/utils.py`. (Although einops still has a similar warning--we'll have to bump that to a more recent version at some point.)

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -128,6 +128,13 @@ docs = [
   "mike>=1.1.2",
 ]
 
+[tool.uv]
+override-dependencies = [
+  # replace unmaintained pydub, which gradio requests, by pydub_ng
+  "pydub; python_version < '0'",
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
ty @joanise

PR #839, opened 2026-07-16
Dev.ej/remove run test

joanise:
### PR Goal? 

We now run tests using `pytest`, and our old `run_tests.sh` script, which served to make it easier to run the tests, has little remaining value, and just creates a duplicate test driver we need to maintain.

It is currently stale. Instead of fixing it, this PR removes it.

### Feedback sought? 

Just confirmation we no longer need run_tests.py.

### Priority? 

low

### Tests added? 

n/a

### Confidence? 

high


roedoejet (approved):
lgtm

PR #852, opened 2026-08-17
Fix vocoder matching

roedoejet:
fix filenames for vocoder matching

### PR Goal? 

Korin reported that vocoder matching was broken (https://docs.everyvoice.ca/stable/guides/finetune/). I confirmed and it appears this was broken by the chunking functionality. This is an important step for FS2 models

### Fixes? 

Issue reported through personal communication.

### Feedback sought? 

Rubber stamping.

### Priority? 

high

### Tests added? 

yes, updated tests to match functionality

### How to test? 

if you synthesize with a filelist, it should use the provided basename (previously it used the old truncate + slugify pipeline)

### Confidence? 

high, I've re-run this and the vocoder matching instructions now work as expected

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/153


joanise (changes requested):
You're killing me with these contentless commit messages and PR titles.

PR #863, opened 2026-08-27
fix: consistently use --config-args in all CLI command

### LINKED ISSUE #861: Inconsistent use of --config and --config-args to (sometimes) mean the same thing
### Bug description

`everyvoice demo` and `everyvoice synthesize` accept `--config` to override specific config values from the CLI.
`everyvoice preprocess` and `everyvoice train` accept `--config-args` to do the same.

Since these behave identically, it would be better UX to make the the same option. I recommend `--config-args`.

Furthermore, `everyvoice g2p` and `everyvoice check text-config` use `--config` to accept the path to a config file.

UX-wise again: we should not have `--config` take on different meanings in different commands.

joanise:
### PR Goal? 

Consistently use `--config-args` to set `config_args` and reserve `--config` for the config file path.

### Fixes? 

Fixes #861

### Priority? 

pre 0.5.0

### How to test? 

```
everyvoice demo -h
everyvoice synthesize from-text -h
```
and see `--config-args` instead of `--config`, along with help text.

### Confidence? 

high

### DIFF
--- a/everyvoice/base_cli/interfaces.py
+++ b/everyvoice/base_cli/interfaces.py
@@ -29,7 +29,10 @@ def load_config_base_command_interface(
     config_file: Annotated[
         Path, typer_file_argument(help="The path to your model configuration file.")
     ],
-    config_args: Annotated[list[str], typer.Option("--config", "-c")] = [],
+    config_args: Annotated[
+        list[str],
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
great, as discussed

PR #864, opened 2026-08-27
fix: preprocess --stats should enable stats, not disable it

### LINKED ISSUE #860: preprocess --stats has counterintuitive negative behaviour
### Bug description

When you call `everyvoice preprocess text-to-spec`, the `--stats` switch has reversed meaning: since the default is `True`, specifying `--stats` or `-S` sets `compute_stats=False` and thus skips the stats. If it's meant to be `--stats` for "on" and `--no-stats` for "off", we need to say:
```
compute_stats: bool = typer.Option(
        True, "-S", "--stats/--no-stats", help="Calculate stats for energy and pitch"
    ),
```
instead of:
```
compute_stats: bool = typer.Option(
        True, "-S", "--stats", help="Calculate stats for energy and pitch"
    ),
```

Caveat: my suggestion defines `--no-stats` but no equivalent short form, since `-S` remains an alias for `--stats`. Typically one would do lowercase=on/uppercase=off but we already use `-s` for steps so that won't work here.

### How to reproduce the bug

```shell
everyvoice preprocess text-to-spec config/everyvoice-text-to-spec.yaml --stats
```
does not create `preprocessed/stats.json`.

joanise:
https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/157

Fixes #860


roedoejet (approved):
thanks!

PR #866, opened 2026-08-27
docs: add missing vocoder-path option to synthesize example

joanise:
### PR Goal? 

I did a quick review of the current online documentation to make sure it was in sync with the software with all the changes we've been doing. You've already done a great job of it, all I found that had to be fixed in an error is fs2 synthesis that's clearly been there for a while: the `--vocoder-path` option is required to output wav.

### Fixes? 

Slightly wrong documentation.

### Feedback sought? 

regular review

### Priority? 

before 0.5.0

### Confidence? 

high


roedoejet (approved):
another good catch. I wonder if we can automate these at all

PR #868, opened 2026-08-28
Support phone-based model inference for StyleTTS2

roedoejet:
### PR Goal? 

When I tested the demo, I only tested it with character based models, but this was only less than half of the issue. For inference with phone-based models we needed to follow the FS2 implementation which is:

- even for monolingual models, a lang2id needs to be defined on the model
- that lang code informs the text processing which, if passed characters, will look in g2p_engines for a g2p engine with the same lang code before inference
- I added this, and also added the chunking while I was at it
- lastly, I added a toggle in the demo to say whether the input text is phones or characters since we might want to input either in the demo.
- I added some debugging comments to address this going forward so we can see in the logs how g2p is processing the text into tokens.

Development was assisted with Claude Code Sonnet 5, reviewed by me.

### Feedback sought? 

code review/sanity - is there anything else I'm forgetting here wrt things that are needed for inference?

### Priority? 

before 0.5.0. :) 

### Confidence? 

medium-high (tested with my pre-trained checkpoints)

### Related PRs? 

https://github.com/EveryVoiceTTS/StyleTTS2/pull/27

### DIFF
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -698,7 +698,7 @@ def _peek_model_class(checkpoint_path: Path) -> str:
     return ckpt.get("model_info", {}).get("name", "")
 
 
-def _peek_text_representation(checkpoint_path: Path) -> str:
+def peek_text_representation(checkpoint_path: Path) -> str:
     """Load a checkpoint header and return the model's trained text representation.
 
     Returns "" (safe default: callers should treat this as "hide any input-type
--- a/everyvoice/cli.py
+++ b/everyvoice/cli.py
@@ -698,6 +698,33 @@ def _peek_model_class(checkpoint_path: Path) -> str:
     return ckpt.get("model_info", {}).get("name", "")
 
 
+def _peek_text_representation(checkpoint_path: Path) -> str:
+    """Load a checkpoint header and return the model's trained text representation.
+
+    Returns "" (safe default: callers should treat this as "hide any input-type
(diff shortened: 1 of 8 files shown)


everyvoice/cli.py
```diff
@@ -698,6 +698,33 @@ def _peek_model_class(checkpoint_path: Path) -> str:
     return ckpt.get("model_info", {}).get("name", "")
 
 
+def _peek_text_representation(checkpoint_path: Path) -> str:
```
joanise: Since you're importing it elsewhere, it's not an internal function: it should not starts with `_`.

everyvoice/demo/app.py
```diff
@@ -483,6 +519,7 @@ def synthesize_audio_styletts2(
     acoustic_blend: float,
     prosody_blend: float,
     language: "str | None" = None,  # selected language code, or None for monolingual checkpoints
+    text_representation: "DatasetTextRepresentation | str" = DatasetTextRepresentation.characters,  # from a live Radio, or a fixed gr.State when the selector is hidden
```
joanise: I don't understand why this type declaration is quoted here (and several others are too).

I know one can use quoted type declaration to use types that aren't defined when they're first encountered, but here things are already imported and the types are visible, so why quote them? (Is there some other use case for such quotes I'm not familiar with?)

joanise (commented):
Initial comments

joanise (approved):
Looks good to me (once CI passes).

PR #869, opened 2026-08-28
Refactor StyleTTS2Module to StyleTTS2

roedoejet:
This is a refactor for the PyTorch Lightning based Module suffixed StyleTTS2Module class (and other related `module` named variables) to StyleTTS2 (and `model` named variables) which is more in-line with other EveryVoice models. It also retains backward compat with StyleTTS2Module trained checkpoints.

Assisted by Claude Sonnet 5

Related:

https://github.com/EveryVoiceTTS/StyleTTS2/pull/28

### DIFF
--- a/everyvoice/tests/test_demo.py
+++ b/everyvoice/tests/test_demo.py
@@ -454,6 +454,9 @@ class TestDemo:
             fake_ckpt = tmpdir / "styletts2.ckpt"
             torch.save(
                 {
+                    # "StyleTTS2Module" is the pre-rename class name; older
+                    # checkpoints carry it and must still dispatch to the
+                    # StyleTTS2 demo path.
                     "model_info": {"name": "StyleTTS2Module"},
                     "hyper_parameters": {"mode": "second", "config": {}},
                     "state_dict": {},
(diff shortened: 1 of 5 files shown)


everyvoice/tests/test_demo.py
```diff
@@ -454,6 +454,9 @@ def test_demo_dispatch_styletts2_rejects_vocoder_flag(self):
             fake_ckpt = tmpdir / "styletts2.ckpt"
             torch.save(
                 {
+                    # "StyleTTS2Module" is the pre-rename class name; older
+                    # checkpoints carry it and must still dispatch to the
+                    # StyleTTS2 demo path.
                     "model_info": {"name": "StyleTTS2Module"},
```
joanise, line 460: Good idea to exercise this recognition of the old name in testing.

joanise (approved):
This looks good to me. Thanks for changing it to model and the class name to just `StyleTTS2`, that'll make a much nicer intuitive and parallel structure with our other models.

Approving now with the assumption that you'll test with old checkpoints to confirm it works.

And thanks for making this its own PR, it's good to leave a separate record for this refactoring work.

PR #876, opened 2026-09-02
docs: add redirects for all pages to stable/

### LINKED ISSUE #761: We should set up redirects for documentation
### Bug description

https://docs.everyvoice.ca/guides should redirect to https://docs.everyvoice.ca/stable/guides for example.

joanise:
### PR Goal? 

Add redirects for all pages foo/bar/ that exist in stable/ from the root, i.e., docs.everyvoice.ca/foo/bar will now redirect to docs.everyvoice.ca/stable/foo/bar.

For example, https://docs.everyvoice.ca/guides/ already redirects to stable/guides (I pushed that one to test my solution), this PR will add the redirect for the rest of our documentation pages.

### Unexpected base branch warning

Note: this PR merges into gh-pages directly.

### Fixes? 

Fixes #761

### Feedback sought? 

Limitations of this solution:
 - we don't control the server, so we cannot set up server-side redirects, which would have been nicer
 - because of that, it's one redirect page per page we want redirected... not DRY
 - maintenance debt: when we add or change pages, we'll have to remember to update the redirects
 - the script version for each case calculates the redirect, but there are still three repeated occurrence of the target path in each redirect page, which is not ideal but probably not avoidable.

Contrastive solution:
 - maybe instead of these redirects, we could use symlinks guides->stable/guides, etc. That would not have to be recursive, but I'd have to test if a symlink that changes the URL depth works.
 - this would require symlinks for guides, assets, references, install, possibly all folders that exist under stable/, but we would not need to create anything recursive because the symlinks will automatically handle that.

Q: which solution do you prefer?

### Priority? 

low

### Tests added? 

no

### How to test? 

Click on https://docs.everyvoice.ca/guides/ to see that the solution works.

You can also test the contrastive (symlink) solution by clicking on https://docs.everyvoice.ca/reference/ (where only the subpage of reference works because I only created the symlinks for assets and references at this point, for testing purposes)

### Confidence? 

 - high that this works
 - medium that this is the right thing to do

### Version change? 

no

### Related PRs? 

no

### DIFF
--- a/guides/background/index.html
+++ b/guides/background/index.html
@@ -4,7 +4,7 @@
   <meta charset="utf-8">
   <title>Redirecting</title>
   <noscript>
-    <meta http-equiv="refresh" content="1; url=/stable/guides/" />
+    <meta http-equiv="refresh" content="1; url=/stable/guides/background/" />
   </noscript>
   <script>
     window.location.replace(
@@ -13,6 +13,6 @@
   </script>
 </head>
 <body>
-  Redirecting to <a href="/stable/guides/">/stable/guides/</a>...
+  Redirecting to <a href="/stable/guides/background/">/stable/guides/background/</a>...
 </body>
 </html>
--- a/guides/custom/index.html
+++ b/guides/custom/index.html
@@ -4,7 +4,7 @@
   <meta charset="utf-8">
   <title>Redirecting</title>
   <noscript>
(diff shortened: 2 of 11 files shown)


roedoejet:
closing in favour of https://github.com/EveryVoiceTTS/EveryVoice/pull/878

PR #880, opened 2026-09-08
Dev.ej/refactor cli api

joanise:


joanise:
Changed my mind, I won't want to remove merge_args. The intellisense for the CLI functions is pointless, it's for the API functions I'll want to optimize it.

PR #884, opened 2026-09-15
Dev.ej/safe refactors

joanise:
### PR Goal? 

Refactoring while working on the CLI/API refactoring, but that don't require it, for separate and simpler review.

 - replace `typing.Union` by ` | ` since we're Python >= 3.10
 - `update_config_from_cli_args` really works by side effects, modifying it's input, but it confusingly also returned it. I prefer following Trey Hunner's advice to either return a new object or mutate the input, but not combining both -- he recommends using method chaining only when each method returns a modified copy, and I think that approach makes cleaner code.

### Feedback sought? 

regular review

### Priority? 

low

### Tests added? 

already covered

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/163
https://github.com/EveryVoiceTTS/StyleTTS2/pull/33


roedoejet (approved):
looks good! and agreed about the modification/return bit

PR #890, opened 2026-09-18
Bump all pre commit hooks to their latest version

joanise:
### PR Goal? 

The real motivation here is that mypy has seen nice speed improvements in recent versions, and I would like to benefit from them. That also makes it a good time to bump all our pre-commit hooks in ev and all submodules.

### Feedback sought? 

probably just rubber stamping

### Priority? 

low

### Tests added? 

n/a

### How to test? 

`prek run --all-files` in the root and all submodules. Should install updated hooks, but on subsequent runs the mypy step runs significantly faster than before.

### Confidence? 

high

### Version change? 

nope

### Related PRs? 

https://github.com/EveryVoiceTTS/HiFiGAN_iSTFT_lightning/pull/69
https://github.com/EveryVoiceTTS/FastSpeech2_lightning/pull/168
https://github.com/EveryVoiceTTS/StyleTTS2/pull/38
https://github.com/EveryVoiceTTS/wav2vec2aligner/pull/33


roedoejet (approved):
looks good! works good. thanks

